PR #22, opened 2020-07-23
README: correct mistake in GPG key decryption explanation

mweinelt:
This looks like a mixup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ If you get:
 ssh-to-pgp: failed to parse private ssh key: ssh: this private key is passphrase protected
 ```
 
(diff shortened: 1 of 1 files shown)


Mic92:
Thanks!

README.md
```diff
@@ -141,7 +141,7 @@ If you get:
 ssh-to-pgp: failed to parse private ssh key: ssh: this private key is passphrase protected
 ```
 
-then your ssh key is encrypted with your password and you need to create a encrypted copy temporarily:
+then your ssh key is encrypted with your password and you need to create an uencrypted copy temporarily:
```
mweinelt, line 144: Awesome, now we traded the explanation mixup for a typo.

Mic92: fixed.

PR #28, opened 2020-08-02
drone-ci: switch to scratch image

Mic92:


Mic92:
Does not work with drone-ci

PR #32, opened 2020-08-08
sops-install-secrets: symlinkSecret: set uid/gid (with Fchownat)

### LINKED ISSUE #30: sops.secrets."testFile" => owner does not propogate to the path symlink
The owner I marked for a given secret is not applied to the symlink:

```nix
{
 sops.secrets."cachix.dhall" = {
      owner = "cole";
      path = "/home/cole/.config/cachix/cachix.dhall";
      format = "binary";
      sopsFile = ./cachix.dhall.sops;
    };
}
```

```console
> ls -al ~/.config/cachix/cachix.dhall
lrwxrwxrwx 1 root root 25 Aug  7 01:40 /home/cole/.config/cachix/cachix.dhall -> /run/secrets/cachix.dhall
```

colemickens:
To keep the existing state machine logic to ensure symlink... this also checks existing symlinks to make sure they're owner/group are correct.

Fixes #30

### DIFF
--- a/pkgs/sops-install-secrets/main.go
+++ b/pkgs/sops-install-secrets/main.go
@@ -158,7 +158,7 @@ func symlinkSecret(targetFile string, secret *secret) error {
 				return fmt.Errorf("Cannot create symlink '%s': %w", secret.Path, err)
 			}
 			if err := secureSymlinkChown(secret.Path, targetFile, secret.owner, secret.group); err != nil {
-				return fmt.Errorf("Cannot secure symlink '%s': %w", secret.Path, err)
+				return fmt.Errorf("Cannot chown symlink '%s': %w", secret.Path, err)
 			}
 			return nil
 		} else if err != nil {
--- a/pkgs/sops-install-secrets/main.go
+++ b/pkgs/sops-install-secrets/main.go
@@ -15,6 +15,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"syscall"
 
 	"github.com/Mic92/sops-nix/pkgs/sshkeys"
 
@@ -102,6 +103,28 @@ type appContext struct {
 	checkMode   CheckMode
 }
 
+func secureSymlinkChown(symlinkToCheck, expectedTarget string, owner, group int) error {
+	fd, err := unix.Open(symlinkToCheck, unix.O_CLOEXEC|unix.O_PATH|unix.O_NOFOLLOW, 0)
+	if err != nil {
+			return fmt.Errorf("Failed to open %s: %w", symlinkToCheck, err)
+	}
+	defer unix.Close(fd)
+
+	buf := make([]byte, len(expectedTarget) + 1) // oversize by one to detect trunc
+	n, err := unix.Readlinkat(fd, "", buf)
+	if err != nil {
+		return fmt.Errorf("couldn't readlinkat %s", symlinkToCheck)
+	}
+	if n > len(expectedTarget) || string(buf[:n]) != expectedTarget  {
+		return fmt.Errorf("symlink %s does not point to %s", symlinkToCheck, expectedTarget)
+	}
+	err = unix.Fchownat(fd, "", owner, group, unix.AT_EMPTY_PATH)
+	if err != nil {
+		return fmt.Errorf("cannot change owner of '%s' to %d/%d: %w", symlinkToCheck, owner, group, err)
+	}
+	return nil
+}
+
 func readManifest(path string) (*manifest, error) {
 	file, err := os.Open(path)
 	if err != nil {
@@ -116,12 +139,26 @@ func readManifest(path string) (*manifest, error) {
 	return &m, nil
 }
 
+func linksAreEqual(linkTarget, targetFile string, info os.FileInfo, secret *secret) bool {
+	validUG := true;
+	if stat, ok := info.Sys().(*syscall.Stat_t); ok {
+		validUG = validUG && int(stat.Uid) == secret.owner
+		validUG = validUG && int(stat.Gid) == secret.group
+	} else {
+		panic("Failed to cast fileInfo Sys() to *syscall.Stat_t. This is possibly an unsupported OS.")
+	}
+	return linkTarget == targetFile && validUG
+}
+
 func symlinkSecret(targetFile string, secret *secret) error {
 	for {
 		stat, err := os.Lstat(secret.Path)
 		if os.IsNotExist(err) {
 			if err := os.Symlink(targetFile, secret.Path); err != nil {
 				return fmt.Errorf("Cannot create symlink '%s': %w", secret.Path, err)
+			}
+			if err := secureSymlinkChown(secret.Path, targetFile, secret.owner, secret.group); err != nil {
+				return fmt.Errorf("Cannot secure symlink '%s': %w", secret.Path, err)
 			}
 			return nil
 		} else if err != nil {
@@ -133,7 +170,7 @@ func symlinkSecret(targetFile string, secret *secret) error {
 				continue
 			} else if err != nil {
 				return fmt.Errorf("Cannot read symlink '%s': %w", secret.Path, err)
-			} else if linkTarget == targetFile {
+			} else if linksAreEqual(linkTarget, targetFile, stat, secret) {
 				return nil
 			}
 		}
@@ -154,7 +191,7 @@ func symlinkSecrets(targetDir string, secrets []secret) error {
 			return fmt.Errorf("Cannot create parent directory of '%s': %w", secret.Path, err)
 		}
 		if err := symlinkSecret(targetFile, &secret); err != nil {
-			return err
+			return fmt.Errorf("Failed to symlink secret '%s': %w", secret.Path, err)
 		}
 	}
 	return nil


colemickens:
Hm. It was easier to make the `Lchown` work.

Why does this need `fchown`?

The `Lchown` docs say: 

>Lchown changes the numeric uid and gid of the named file. If the file is a symbolic link, it changes the uid and gid of the link itself. If there is an error, it will be of type *PathError.

colemickens:
Hm... https://man7.org/linux/man-pages/man2/fchown.2.html

pkgs/sops-install-secrets/main.go
```diff
@@ -116,13 +117,25 @@ func readManifest(path string) (*manifest, error) {
 	return &m, nil
 }
 
+func linksAreEqual(linkTarget, targetFile string, info os.FileInfo, secret *secret) bool {
+	validUG := true;
+	if stat, ok := info.Sys().(*syscall.Stat_t); ok {
```
Mic92, line 122: I think this should panic in the `!ok` case. This way if we ever target a new operating system this error does not get hidden.

pkgs/sops-install-secrets/main.go
```diff
@@ -116,13 +117,25 @@ func readManifest(path string) (*manifest, error) {
...
+}
+
 func symlinkSecret(targetFile string, secret *secret) error {
 	for {
 		stat, err := os.Lstat(secret.Path)
 		if os.IsNotExist(err) {
 			if err := os.Symlink(targetFile, secret.Path); err != nil {
 				return fmt.Errorf("Cannot create symlink '%s': %w", secret.Path, err)
 			}
+			if err := os.Lchown(secret.Path, secret.owner, secret.group); err != nil {
```
Mic92, line 136: The reason I wanted to use fchown is that a potential attacker could replace the symlink with a file and gaining access to a file that was not supposed to be read.

To circumvent that the following approach makes sure we only change the owner of our own symlink:

1. fd = open() with O_PATH set -> this allows to open symlinks
2. readlinkat(fd, ""); the "" is a special case which allows to operate on file descriptors
3. check if the result of readlinkat is as expected
4. fchown(fd, ...)

I have not checked if the go stdlib exposes all these but `golang.org/x/sys/unix` shoul

colemickens: I just want to make sure I understand. We create the symlink and then immediately `Lchown` it. It's the only time Lchown would be called. So the attacker in this scenario can pause execution, or time-attack fast enough to switch the symlink after we create, and before we symlink?

I do think I understand that the method you described will operate on the exact symlink-fd and let us ensure the target is valid before we act on it.

Since this always happens when we create the symlink, is it easier to create the symlink initially with the correct owner and group?

Mic92: Creating a symlink initially with the correct permission is possible too, but I am not sure it is easier. I think what would work is to use `setfsuid`/`setfsguid` to run the system call with the user's owner/group. However this can fail if the user does not have actually write access to the directory you are trying to write. TOCTOU attacks in the filesystem are certainly possible - I have implemented them myself - There are beginner examples in CTFs.

Mic92: At the end of the day do we want to allow users to interfere with system provisioned symlinks?

colemickens:
@Mic92 Do you know why this is hard to find documentation? I expected to find "lchown considered harmful" or something like that.

Also, this works. The `Fchownat` and `unix.AT_EMPTY_PATH|unix.AT_SYMLINK_NOFOLLOW` were crucial.

pkgs/sops-install-secrets/main.go
```diff
@@ -102,6 +103,30 @@ type appContext struct {
...
+			return fmt.Errorf("Failed to open %s: %w", symlinkToCheck, err)
+	}
+	defer unix.Close(fd)
+
+	buf := make([]byte, len(expectedTarget) + 1)
+	n, err := unix.Readlinkat(fd, "", buf)
+	if err != nil {
+		return fmt.Errorf("couldn't readlinkat %s", symlinkToCheck)
+	}
+	actualTarget := string(buf[:n]) // chop off null byte
```
Mic92, line 118: According to `readlink(2)` there is no terminating null byte appended.

colemickens: I'll look into it, but without it there was a \x00 that was causing the string comparison to fail.

pkgs/sops-install-secrets/main.go
```diff
@@ -102,6 +103,30 @@ type appContext struct {
...
 
+func secureSymlinkChown(symlinkToCheck, expectedTarget string, owner, group int) error {
+	fd, err := unix.Open(symlinkToCheck, unix.O_CLOEXEC|unix.O_PATH|unix.O_NOFOLLOW, 0)
+	if err != nil {
+			return fmt.Errorf("Failed to open %s: %w", symlinkToCheck, err)
+	}
+	defer unix.Close(fd)
+
+	buf := make([]byte, len(expectedTarget) + 1)
+	n, err := unix.Readlinkat(fd, "", buf)
```
Mic92, line 114: If n is larger than `symlinkToCheck` than there might be truncation happening. This was the actual reason I added + 1 in https://github.com/Mic92/sops-nix/pull/36/files#diff-60efc9b29f1818e86a75e498ef70348aR136
In https://github.com/Mic92/sops-nix/pull/36/files#diff-60efc9b29f1818e86a75e498ef70348aR138
I check if the same: https://github.com/Mic92/sops-nix/pull/36/files#diff-60efc9b29f1818e86a75e498ef70348aR138

pkgs/sops-install-secrets/main.go
```diff
@@ -102,6 +103,30 @@ type appContext struct {
...
+	if err != nil {
+		return fmt.Errorf("couldn't readlinkat %s", symlinkToCheck)
+	}
+	actualTarget := string(buf[:n]) // chop off null byte
+
+	if !strings.EqualFold(expectedTarget, actualTarget) {
```
Mic92, line 120: Linux filesystems should not perform unicode normalisation. So why not compare bytes here?

pkgs/sops-install-secrets/main.go
colemickens: There's a relevant note at the bottom: https://www.gnu.org/software/gnulib/manual/html_node/readlinkat.html

pkgs/sops-install-secrets/main.go
colemickens: Mostly a leftover from troubleshooting the aforementioned issue, I'll switch this back to a regular `==`.

pkgs/sops-install-secrets/main.go
Mic92: The `\x00` is because your byte buffer is 1 byte longer than your string. You need to compare `buf[:n]` to your string.

colemickens: never mind, I think I understand from the other thread.

pkgs/sops-install-secrets/main.go
colemickens: I think I get it now. I'll update this a smidge and fix the comment.

pkgs/sops-install-secrets/main.go
Mic92: so your code is actually correct but maybe fix the comment because it is misleading. In fact you could cast the input string to a byte slice as well and compare those to make it very explicit that you are just comparing bytes.

colemickens:
Okay there we go, now it looks more like your diff. (If you want to just take the flags and put them on your diff for expediency that's fine of course). Thanks for the fast review.

Mic92:
> @Mic92 Do you know why this is hard to find documentation? I expected to find "lchown considered harmful" or something like that.
> 
> Also, this works. The `Fchownat` and `unix.AT_EMPTY_PATH|unix.AT_SYMLINK_NOFOLLOW` were crucial.

Maybe if you search for `lchown` and `toctu` you will find more results.

pkgs/sops-install-secrets/main.go
```diff
@@ -102,6 +103,28 @@ type appContext struct {
...
+
+	buf := make([]byte, len(expectedTarget) + 1) // oversize by one to detect trunc
+	n, err := unix.Readlinkat(fd, "", buf)
+	if err != nil {
+		return fmt.Errorf("couldn't readlinkat %s", symlinkToCheck)
+	}
+	if n > len(expectedTarget) || string(buf[:n]) != expectedTarget  {
+		return fmt.Errorf("symlink %s does not point to %s", symlinkToCheck, expectedTarget)
+	}
+	err = unix.Fchownat(fd, "", owner, group, unix.AT_EMPTY_PATH|unix.AT_SYMLINK_NOFOLLOW)
```
Mic92, line 121: Since our file descriptor already points to right inode, we can skip `AT_SYMLINK_NOFOLLOW` here:

```suggestion
	err = unix.Fchownat(fd, "", owner, group, unix.AT_EMPTY_PATH)
```

It does not make a function difference here but at least make it clear that we understood the manpage.

pkgs/sops-install-secrets/main.go
```diff
@@ -102,6 +103,28 @@ type appContext struct {
...
 }
 
+func secureSymlinkChown(symlinkToCheck, expectedTarget string, owner, group int) error {
+	fd, err := unix.Open(symlinkToCheck, unix.O_CLOEXEC|unix.O_PATH|unix.O_NOFOLLOW, 0)
+	if err != nil {
+			return fmt.Errorf("Failed to open %s: %w", symlinkToCheck, err)
+	}
+	defer unix.Close(fd)
+
+	buf := make([]byte, len(expectedTarget) + 1) // oversize by one to detect trunc
```
Mic92, line 113: I like that truncation is truncated here.

pkgs/sops-install-secrets/main.go
```diff
@@ -102,6 +103,28 @@ type appContext struct {
...
+	n, err := unix.Readlinkat(fd, "", buf)
+	if err != nil {
+		return fmt.Errorf("couldn't readlinkat %s", symlinkToCheck)
+	}
+	if n > len(expectedTarget) || string(buf[:n]) != expectedTarget  {
+		return fmt.Errorf("symlink %s does not point to %s", symlinkToCheck, expectedTarget)
+	}
+	err = unix.Fchownat(fd, "", owner, group, unix.AT_EMPTY_PATH)
+	if err != nil {
+		return fmt.Errorf("cannot Fchownat %q: %w", symlinkToCheck, err)
```
Mic92, line 123:
```suggestion
		return fmt.Errorf("cannot change owner of %q to %d:%d: %w", symlinkToCheck, owner, group, err)
```

Mic92: I have not seen `%q` so far should it not be `%s`?

colemickens:
After all of this, this isn't quite enough anyway. The permissions on `/run/secrets{,.d}` are also too restrictive, even though the symlink and secret are correct, the parent directories aren't executable so my regular user can't access them.

Mic92:
> After all of this, this isn't quite enough anyway. The permissions on `/run/secrets{,.d}` are also too restrictive, even though the symlink and secret are correct, the parent directories aren't executable so my regular user can't access them.

This was adapted from the existing nixops design. I was wondering if I should remove the `keys` group requirement completely. It requires all services to be in the `keys` group too. After all creating secrets with the right permission should be sufficient.

pkgs/sops-install-secrets/main.go
colemickens:
```
%s	the uninterpreted bytes of the string or slice
%q	a double-quoted string safely escaped with Go syntax
```

Consistent is probably best.... I took your suggestion verbatim and applied it.

colemickens:
Oh right, I don't know how I didn't notice this while staring at the permissions last night. I added `keys` to my `extraGroups`, opened a new shell with `newgrp keys` and then was able to access the secret, as expected. Seems fine to me, I think?

pkgs/sops-install-secrets/main.go
```diff
@@ -116,13 +139,27 @@ func readManifest(path string) (*manifest, error) {
...
+
 func symlinkSecret(targetFile string, secret *secret) error {
 	for {
 		stat, err := os.Lstat(secret.Path)
 		if os.IsNotExist(err) {
 			if err := os.Symlink(targetFile, secret.Path); err != nil {
 				return fmt.Errorf("Cannot create symlink '%s': %w", secret.Path, err)
 			}
+			if err := secureSymlinkChown(secret.Path, targetFile, secret.owner, secret.group); err != nil {
+				return fmt.Errorf("Cannot secure symlink '%s': %w", secret.Path, err)
```
Mic92, line 161:
```suggestion
				return fmt.Errorf("Cannot chown symlink '%s': %w", secret.Path, err)
```

PR #39, opened 2020-10-26
Add option for choosing which activation method to use

zaninime:
I just tested this on my machine and it seems to perform fine. The default behavior is set to the "old" way, so it's not a breaking change for existing users. Opting-in should be as simple as `sops.activationMethod = "systemd";`

I used an enum for now, but the alternative would be to use a boolean, ie. `useSystemd`. I'm happy to change it.

### DIFF
--- a/modules/sops/default.nix
+++ b/modules/sops/default.nix
@@ -87,6 +87,11 @@ let
     sops-install-secrets -check-mode=${if cfg.validateSopsFiles then "sopsfile" else "manifest"} ${manifest}
     cp ${manifest} $out
   '';
+
+  setupScript = ''
+      echo setting up secrets...
+      ${optionalString (cfg.gnupgHome != null) "SOPS_GPG_EXEC=${pkgs.gnupg}/bin/gpg"} ${sops-install-secrets}/bin/sops-install-secrets ${checkedManifest}
+  '';
 in {
   options.sops = {
     secrets = mkOption {
@@ -140,6 +145,15 @@ in {
         This option must be explicitly unset if <literal>config.sops.sshKeyPaths</literal>.
       '';
     };
+
+    activationMethod = mkOption {
+      type = types.enum [ "script" "systemd" ];
+      default = "script";
+      description = ''
+        Which method to use for setting up secrets. Use `script` for an
+        activation script, and `systemd` for a systemd unit.
+      '';
+    };
   };
   config = mkIf (cfg.secrets != {}) {
     assertions = [{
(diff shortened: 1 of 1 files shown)


modules/sops/default.nix
```diff
@@ -140,6 +145,15 @@ in {
         This option must be explicitly unset if <literal>config.sops.sshKeyPaths</literal>.
       '';
     };
+
+    activationMethod = mkOption {
+      type = types.enum [ "script" "systemd" ];
+      default = "script";
+      description = ''
+        Which method to use for setting up secrets. Use `script` for an
+        activation script, and `systemd` for a systemd unit.
```
Mic92, line 154: I think `systemd` is the way to go to use all features of sops-nix. I am thinking about deprecating the activation script.

zaninime: Should I change this PR to reflect this? I'd be happy to do it.

Mic92: I hope I get to hack next weekend again on sops-nix: https://github.com/systemd/systemd/pull/16568#issuecomment-720072151
I prepared systemd socket activation and with this PR https://github.com/systemd/systemd/pull/17510 I should be able to implement it properly and ready for networked KMS.

Mic92: To be more explicit with socket-based secret description it is easier to make systemd wait for secrets to become available without having to specify explicit dependencies. This was my major concern with having sops-nix running as a systemd service instead of activation script. I hope to be able to implement this soonish.

aanderse:
Hi @Mic92 :wave:

Sorry to pull up an old thread but...

> I think systemd is the way to go to use all features of sops-nix. I am thinking about deprecating the activation script method.

Seeing as this PR was never closed is there still any interest in switching to a `systemd` service instead of activation scripts?

cransom:
I have a particular use case where if sops-nix were to switch to a systemd service, it would open up secrets usage for short lived systems. Right now, sops-nix assumes GPG/age/ssh keys and it requires that key material exist and a deploy/nixos-rebuild. sops, if using AWS KMS (or any other KMS), will decrypt without that material at run time if roles and access to the correct KMS key is granted. I'd be happy to work a PR if this functionality is thought useful to the project.

k3d3:
I'm interested in this in particular because it would allow me to use my machine's TPM as a root of trust, rather than my host SSH key.

snylonue:
What's the current status of the pr?

Mic92:
I believe this is solved by https://github.com/Mic92/sops-nix/pull/855

If I am missing something, please let me know.

PR #67, opened 2021-01-26
Fix flake build

edibopp:
The flake build did not work on the latest master, specifically `nix flake show` and `nix build .#unit-tests`. I fixed the issues and added basic checks to the CI build to prevent regressions.

Checks are not configured yet, so `nix flake check -L` does not work.


Mic92:
Thanks!

PR #70, opened 2021-01-27
add test data

Mic92:


Mic92:
I don't think I am motivated enough to fix gpg-related issues.

PR #74, opened 2021-02-01
pkgs: don't reference deprecated stdenv.lib

lovesegfault:
`stdenv.lib` has been deprecated in favor of using `lib` directly.


lovesegfault:
The CI failure seems to be caused by too strict a timeout?

PR #77, opened 2021-02-01
systemd secret support

Mic92:


Mic92:
Background: https://github.com/systemd/systemd/pull/16568

Mic92:
Systemd's LoadCredentials is in a bit weird state i.e. it does not work for service reloads.
It also seems that nixos is the main user of it. For now I rather would stick to activation phase which works well for on-device keys.

PR #81, opened 2021-02-08
Fix sops-pgp-hook erroring in a strict shell

### LINKED ISSUE #80: sops-pgp-hook evaluation fails in a strict shell (set -euo pipefail)
While testing things in a local repository, I discovered with `direnv` that enabling the `strict_env` option, which executes the hook with `set -euo pipefail`, would fail to evaluate the hook and exit.

I didn't have the variable below exposed so the script would exit with `$VARIABLE_NAME: unbound variable`

https://github.com/Mic92/sops-nix/blob/4a8cb2bd5051cdce7767bc0ea817aedb3896462d/pkgs/sops-pgp-hook/sops-pgp-hook.bash#L14
https://github.com/Mic92/sops-nix/blob/4a8cb2bd5051cdce7767bc0ea817aedb3896462d/pkgs/sops-pgp-hook/sops-pgp-hook.bash#L21

Changing the logic to `"${VARIABLE_NAME-}"` would prevent errors and only assign the empty string if the variable is undefined.

berbiche:
Fixes #80

### DIFF
--- a/pkgs/sops-pgp-hook/sops-pgp-hook.bash
+++ b/pkgs/sops-pgp-hook/sops-pgp-hook.bash
@@ -5,20 +5,20 @@ _sopsAddKey() {
   fpr=$(@gpg@ --with-fingerprint --with-colons --show-key "$key" \
          | awk -F: '$1 == "fpr" { print $10; exit }')
   if [[ $fpr != "" ]]; then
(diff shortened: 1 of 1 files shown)


Mic92:
Thanks!

PR #84, opened 2021-02-22
readme: add `nix shell` for explicit example

colemickens:
Some forgetful folks use `sops-nix` a lot but keep forgetting the fastest way to onboard a new machine. This is the most minimal option for getting to `ssh-to-pgp` that doesn't rely on users having `NIX_PATH` set).


colemickens:
Alternatively, the flake could wire up such that this would work, I think: `nix run github:Mic92/sops-nix#ssh-to-gpg -o key.asc`

Mic92:
I already added ssh-to-pgp some time ago to nixpkgs: https://github.com/Mic92/sops-nix/pull/78

Issue #48, opened 2020-11-24
"/nix/store/... is not in the nix store"


ncfavier:
I'm doing something like

```nix
{
  sops.secrets.foo.sopsFile = "${secretsPath}/foo"; # where secretsPath = "/nix/store/..."
}
```

but this fails because of this check:

https://github.com/Mic92/sops-nix/blob/b0d764fccdd35588ebda2907911099198f73c89a/modules/sops/default.nix#L154

I *could* use `secretsPath + "/foo"` instead, but I feel like I shouldn't have to. Is there a better way to check whether a path is in the Nix store?

Mic92:
One could check if it starts with `builtins.storeDir`.

```nix
nix-repl> builtins.storeDir
"/nix/store"
```

ncfavier:
Maybe with a round-trip through `builtins.toPath` and `builtins.toString` first in order to normalise it.

Mic92:
I don't want to force it to be a path otherwise it might be copied to the nix store without the user intending to do so.

ncfavier:
`toString` converts a path to a string without copying anything to the store.

The manual says `toPath` is deprecated, so it would rather look like:

```nix
lib.hasPrefix builtins.nixStore (toString (/. + sopsFile))
```

Mic92:
Ok. Do you mind open a PR for this?

ncfavier:
Upon further investigation, the scenario I was describing can only work if `nix.useSandbox` is set to `false` (which I don't really want), because otherwise the builder for `checked-manifest.json` doesn't have access to `secretsPath`, even if it's in the store.

So I think it's actually sane to require a path if `validateSopsFile` is enabled.

The error message is a bit confusing though, maybe I'll make a PR to change that.

Mic92:
Thanks!

PR #87, opened 2021-03-03
Improve assertions

ncfavier:
This is the continuation of #48, but I now understand paths and string contexts a little better.

I have
- made the `validateSopsFiles -> isPath sopsFile` assertion also accept strings that are store paths. This has the side effect that `sopsFile` can now refer to arbitrary derivations, so I had to replace `builtins.toFile` with `pkgs.writeText`.
- changed the type of `sopsFile` and `defaultSopsFile` to just `path` (which already accepts both strings and paths).
- merged the two assertions about `gnupgHome` and `sshKeyPaths` into a single one
- moved the `pathExists sopsFile` assertion into `config.assertions`.

The first point is the one that deserves discussion: what I was failing to understand before is that strings that start with `/nix/store` generally have an associated *context*, which my `secretsPath` didn't because I used `toString` on a path. This PR allows `sopsFile = "${secretsPath}/foo";` to work, provided that `secretsPath` is a path and not a string.

### DIFF
--- a/modules/sops/default.nix
+++ b/modules/sops/default.nix
@@ -23,17 +23,14 @@ let
           This option is ignored if format is binary.
         '';
       };
-      path = assert assertMsg (builtins.pathExists config.sopsFile) ''
-          Cannot find path '${config.sopsFile}' set in 'sops.secrets."${config._module.args.name}".sopsFile'
+      path = mkOption {
+        type = types.str;
+        default = "/run/secrets/${config.name}";
+        description = ''
+          Path where secrets are symlinked to.
+          If the default is kept no symlink is created.
         '';
-        mkOption {
-          type = types.str;
-          default = "/run/secrets/${config.name}";
-          description = ''
-            Path where secrets are symlinked to.
-            If the default is kept no symlink is created.
-          '';
-        };
+      };
       format = mkOption {
         type = types.enum ["yaml" "json" "binary"];
         default = cfg.defaultSopsFormat;
@@ -64,7 +61,7 @@ let
         '';
       };
       sopsFile = mkOption {
-        type = types.either types.str types.path;
+        type = types.path;
         default = cfg.defaultSopsFile;
         description = ''
           Sops file the secret is loaded from.
@@ -72,7 +69,7 @@ let
       };
     };
   });
-  manifest = builtins.toFile "manifest.json" (builtins.toJSON {
+  manifest = pkgs.writeText "manifest.json" (builtins.toJSON {
     secrets = builtins.attrValues cfg.secrets;
     # Does this need to be configurable?
     secretsMountPoint = "/run/secrets.d";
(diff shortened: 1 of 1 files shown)


modules/sops/default.nix
```diff
@@ -144,17 +141,17 @@ in {
...
-      message = "Either sops.sshKeyPaths and sops.gnupgHome must be set";
-    }] ++ map (name: let
-      inherit (cfg.secrets.${name}) sopsFile;
-    in {
-      assertion = cfg.validateSopsFiles -> builtins.isPath sopsFile;
-      message = "${sopsFile} is not in the nix store. Either add it to the nix store or set `sops.validateSopsFiles` to false";
-    }) (builtins.attrNames cfg.secrets);
+      assertion = cfg.validateSopsFiles ->
+        builtins.isPath secret.sopsFile ||
+        (builtins.isString secret.sopsFile && hasPrefix builtins.storeDir secret.sopsFile);
```
Mic92, line 152: I find this hard to understand. I assume this is how it will be evaluated?

```suggestion
      assertion = (cfg.validateSopsFiles -> builtins.isPath secret.sopsFile) ||
        (builtins.isString secret.sopsFile && hasPrefix builtins.storeDir secret.sopsFile);
```

modules/sops/default.nix
```diff
@@ -144,17 +141,17 @@ in {
   };
   config = mkIf (cfg.secrets != {}) {
     assertions = [{
-      assertion = cfg.gnupgHome != null -> cfg.sshKeyPaths == [];
-      message = "Configuration options sops.gnupgHome and sops.sshKeyPaths cannot be set both at the same time";
+      assertion = (cfg.gnupgHome == null) != (cfg.sshKeyPaths == []);
+      message = "Exactly one of sops.gnupgHome and sops.sshKeyPaths must be set";
+    }] ++ concatLists (mapAttrsToList (name: secret: [{
+      assertion = builtins.pathExists secret.sopsFile;
```
Mic92, line 147: This might break the use case when you build your system in a CI where the sopsfile are not uploaded to. I think it should be only checked if `validateSopsFiles` is set.

modules/sops/default.nix
ncfavier: `a -> b` is the same as `!a || b`, and `||` is associative, which means it doesn't matter where you put the parentheses. If you want to have them anyway for clarity, I'd put them on the right instead:
```nix
assertion = cfg.validateSopsFiles -> (builtins.isPath secret.sopsFile ||
        (builtins.isString secret.sopsFile && hasPrefix builtins.storeDir secret.sopsFile));
```

Mic92: Yes please.

modules/sops/default.nix
ncfavier: I wrapped the two assertions in `optionals cfg.validateSopsFiles`.

Mic92:
Thanks!

PR #89, opened 2021-03-28
module: Don't set option-value as option-default

Ma27:
When using `documentation.nixos.includeAllModules = true;`, I'd
otherwise have to rebuild the manual on each change since I have my
`defaultSopsFile` in a git-repo with all my other configs.


Mic92:
Thanks!

Issue #53, opened 2020-12-17
Document `age` usage


jtrakk:
Sops [recently got `age` support](https://github.com/mozilla/sops/pull/688). Is it possible to use `age` with sops-nix?

Mic92:
I have not tested it yet. It should be possible by using `SOPS_AGE_KEY_FILE` as described here https://github.com/mozilla/sops/pull/688/files#diff-7b3ed02bc73dc06b7db906cf97aa91dec2b2eb21f2d92bc5caa761df5bbc168fR181
This would require to also update sops here: https://github.com/Mic92/sops-nix/blob/master/go.sum

jtrakk:
Where should I set the `SOPS_AGE_KEY_FILE` variable? If I just do it in the shell where I run `nixos-rebuild`, sops won't see it.

Mic92:
I think we need to extend the module and set it here: https://github.com/Mic92/sops-nix/blob/797e7552776441a018e0a7299b5ba647aed40ffe/modules/sops/default.nix#L160

jtrakk:
For reference the good rev is this PR https://github.com/mozilla/sops/pull/788 that uses the armor format. The earlier one on sops develop branch doesn't work reliably.

Mic92:
I hope I get to implementing converting ssh ed25519 keys to sops/age keys soon similar how it was done for gpg.

jtrakk:
Age natively supports ssh keys, but the Age feature of Sops doesn't include ssh keys.

Mic92:
> Age natively supports ssh keys, but the Sops PR that adds Age doesn't include the ssh keys feature.

That's not an issue. In my implementation for sops-install-secrets I already do the ssh rsa to pgp key conversion. For age that could be applied in the same way.

Mic92:
Proper age support comes here: https://github.com/Mic92/sops-nix/pull/95

PR #95, opened 2021-05-05
Age support

knownunown:
This is my first PR to a Nix-related project, seeking any feedback :)

I've done the following:
- Bumped the `sops` version backing `sops-nix`
- Took a crack at adding some related config options for the NixOS module, `useAge` and `ageKeyFile`
- added some age-related logic to `sops-install-secrets`
- added a `sops-age-hook`, analogous to `sops-pgp-hook` but for age

This version of the patchset doesn't attempt to do any conversions between age keys and SSH keys as mentioned in #53, though I would be open to attempting that. I think that the only pain point in re: lack of SSH key support is that users will have to generate their own age key. Storage and management of those user might end up being a hassle, but the module generates machine-specific age keys just as `sshd` does for host keys. 

At any rate, I still have to document usage of the module with age and add tests, though I wanted to open this PR first to get some eyes on it before proceeding.


pkgs/sops-age-hook/sops-age-hook.bash
```diff
@@ -0,0 +1,25 @@
+_sopsAddKey() {
```
Mic92, line 1: I would suggest not to add a new sops-age-hook but instead encourage users to setup a .sops.yaml instead.

https://github.com/Mic92/dotfiles/blob/master/nixos/.sops.yaml

The reason is that `SOPS_AGE_RECIPIENTS` does not work with multiple secret files.

modules/sops/default.nix
```diff
@@ -134,35 +151,66 @@ in {
...
-        assertion =
-          builtins.isPath secret.sopsFile ||
-          (builtins.isString secret.sopsFile && hasPrefix builtins.storeDir secret.sopsFile);
-        message = "'${secret.sopsFile}' is not in the Nix store. Either add it to the Nix store or set sops.validateSopsFiles to false";
-      }]) cfg.secrets)
-    );
+  config = mkMerge [
+    (mkIf cfg.useAge {
+      system.activationScripts.generate-age-key = ''
+        if [[ ! -f "${cfg.ageKeyFile}" ]]; then
```
Mic92, line 167: How do you feel about deriving this from the ssh host key? NixOS generates ed25515 keys already.
I would also add support for that to this PR if you don't mind.

Mic92: Ok. Just read your PR text. I don't think I will have time to look into this before the weekend.

Mic92:
Thanks for looking into this!

modules/sops/default.nix
knownunown: Yeah, to reiterate, I think it would be pretty useful for user keys, while it would be of limited use for host keys. If we work on the conversion for user keys though, we might as well derive host keys too :)
As for implementation details, the age tool itself seems to make a peculiar distinction between Ed25519 (SSH) and X25519 (age-native) keys. Instead of a pure conversion, they a specific SSH key type and [add some metadata to the key wrapper](https://github.com/FiloSottile/age/blob/85763d390a45b546ad2ac34a5e3e09ccd047c6df/agessh/agessh.go#L223). I'm not sure if that's relevant for our use case, I'll need to take a close look.
Otherwise, The SSH key conversion path looks to be somewhat straightforward, though in a perfect world I would prefer that upstream sops implement native age SSH key support.

pkgs/sops-age-hook/sops-age-hook.bash
knownunown: > The reason is that `SOPS_AGE_RECIPIENTS` does not work with multiple secret files.
Can you clarify what you mean by this? `SOPS_AGE_RECIPIENTS` can encrypt to multiple public keys, I think I have it doing so on my local config.

I do think that we should encourage usage of a `.sops,yaml` though, it's nice and declarative :) I wasn't aware that it was even an option. Seems like i didn't read sops documentation thoroughly enough ...

knownunown: another issue with this hook if we were to keep it -- I don't think it handles multiple public keys in one file correctly, which age seems to encourage

Mic92: Ok. To clarify. It does not work if you are trying to have multiple sops files with different recipients, which you can do with the .sops.yaml approach.

modules/sops/default.nix
Mic92: I asked about native ssh support as well when the feature was added, but they did not want to implement and also the age maintainers that reviewed the sops PR did not encourage it. Also not sure about the ssh metadata. In theory it should not be required.

Mic92: Nevertheless. I think having an activation script generating a new age key is also something, some people prefer. So let's leave it in.

Mic92:
I have not forget about that, it was just that life got in the way.

knownunown:
Yeah, apologies on my end -- it's been on my backburner, pending formalization of secret management for my system configuration. Merged into my branch to hack around a bit more.

Mic92:
There is now a `sops-import-keys-hook`, which could be extended to age as well.

modules/sops/default.nix
```diff
@@ -134,35 +151,66 @@ in {
...
-        message = "Cannot find path '${secret.sopsFile}' set in sops.secrets.${strings.escapeNixIdentifier name}.sopsFile";
-      } {
-        assertion =
-          builtins.isPath secret.sopsFile ||
-          (builtins.isString secret.sopsFile && hasPrefix builtins.storeDir secret.sopsFile);
-        message = "'${secret.sopsFile}' is not in the Nix store. Either add it to the Nix store or set sops.validateSopsFiles to false";
-      }]) cfg.secrets)
-    );
+  config = mkMerge [
+    (mkIf cfg.useAge {
```
Mic92, line 165: I would like to keep the option open to support ssh later:

```suggestion
    (mkIf cfg.generateAgeKey {
```

pkgs/sops-install-secrets/main.go
```diff
@@ -47,6 +47,8 @@ type manifest struct {
 	SymlinkPath       string   `json:"symlinkPath"`
 	SSHKeyPaths       []string `json:"sshKeyPaths"`
 	GnupgHome         string   `json:"gnupgHome"`
+	UseAge            bool     `json:"useAge"`
```
Mic92, line 50: I think `AgeKeyFile` would be enough here.

pkgs/sops-age-hook/sops-age-hook.bash
Mic92: This should be removed now. See https://github.com/Mic92/sops-nix/issues/98

PR #102, opened 2021-07-04
add readonly option sopsFileHash

NickCao:
When secrets are changed, the related services are unlikely to be automatically restarted/reloaded. However `systemd.services.<name>.restartTriggers` can be utilized to achieve that, given that we hash the corresponding sops file beforehand. Another approach can be that we prepend the hash to the path of the secret.

### DIFF
--- a/modules/sops/default.nix
+++ b/modules/sops/default.nix
@@ -70,6 +70,14 @@ let
           Sops file the secret is loaded from.
         '';
       };
+      sopsFileHash = mkOption {
+        type = types.str;
+        readOnly = true;
+        default = if cfg.validateSopsFiles then "${builtins.hashFile "sha256" config.sopsFile}" else "";
+        description = ''
+          Hash of the sops file, useful in systemd.services.<name>.restartTriggers.
+        '';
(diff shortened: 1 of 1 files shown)


modules/sops/default.nix
```diff
@@ -70,6 +70,14 @@ let
           Sops file the secret is loaded from.
         '';
       };
+      sopsFileHash = mkOption {
+        type = types.str;
+        readOnly = true;
+        default = if cfg.validateSopsFiles then "${builtins.hashFile "sha256" config.sopsFile}" else "";
```
Mic92, line 76: Is this different from using `config.sopsFile` directly?

NickCao: `config.sopsFile` might not exist at evaluation time, and can cause evaluation failures.

NickCao: The file path is within the flake source directory in nix store, resulting in unnecessary restarts of services.

jbgi:
Wondering is sopsFileHash should also be available for templates... should I open a PR?

PR #103, opened 2021-07-10
modules/sops: fix manual

Ma27:
* Since 0d957142b6669dcd138fc3518f69d15432c6dcd5 the manual doesn't
  build since `<name>` is interpreted by docbook as (unmatched) XML-tag.
  I decided to use `<xref linkend` as this provides proper linking to
  the referenced option.

* Also, if the module is included on a machine where `sops` isn't used,
  but `documentation.nixos.includeAllModules = true;` is set, the module
  wouldn't evaluate because `config.sopsFile` is referenced in a
  `default`-tag. This is generally an issue since every change to this
  option would trigger a rebuild of the manual anyways.

  See also 94fd200305ceb031380381ae91a1ffabe9f06706 for that.

cc @Mic92 @NickCao

### DIFF
--- a/modules/sops/default.nix
+++ b/modules/sops/default.nix
@@ -8,6 +8,7 @@ let
   secretType = types.submodule ({ config, ... }: {
     config = {
       sopsFile = lib.mkOptionDefault cfg.defaultSopsFile;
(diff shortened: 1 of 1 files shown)


Mic92:
Thanks!

PR #113, opened 2021-09-02
Add actual support for restarting/reloading units, change detection, dry mode

dasJ:


pkgs/sops-install-secrets/main.go
```diff
@@ -509,6 +511,80 @@ func importSSHKeys(keyPaths []string, gpgHome string) error {
...
+			if os.IsNotExist(err) {
+				// File did not exist before
+				restart = append(restart, secret.RestartUnits...)
+				reload = append(reload, secret.ReloadUnits...)
+				continue
+			}
+			return err
+		}
+		defer oldFile.Close()
+		if _, err = io.Copy(hasher, oldFile); err != nil {
```
Mic92, line 538: Is comparing both files byte-wise not easier and faster than calculating sha256 hashes? It's also 100% collision free :)

README.md
```diff
@@ -728,11 +728,6 @@ Each host has a [secrets](https://github.com/Mic92/dotfiles/tree/master/nixos/ev
 
 ## Known limitations
 
-### Restarting systemd services
-
-Right now systemd services are not restarted automatically.
-We want to implement this in future.
-
 ### Initrd secrets
```
Mic92, line 731: Can you extend the `buildkite-agent-builder` example to just have a `restartUnits` value set?

pkgs/sops-install-secrets/main.go
```diff
@@ -509,6 +510,74 @@ func importSSHKeys(keyPaths []string, gpgHome string) error {
...
+			defer f.Close()
+			for _, unit := range list {
+				if _, err = f.WriteString(unit + "\n"); err != nil {
+					return err
+				}
+			}
+		}
+		return nil
+	}
+	if err := writeLines(restart, "/run/systemd/restart-list"); err != nil {
```
Mic92, line 571: Maybe just read both files restart/reload and everything not in the list can be restarted/reloaded from sops-nix itself.
Advantages: This works without changing nixpkgs. If nixpkgs's interface ever changes we might just restart/reload too often, which is not that much of a big deal and gives us some time to fix it.

pkgs/sops-install-secrets/main.go
```diff
@@ -20,36 +21,43 @@ import (
 	"github.com/Mic92/sops-nix/pkgs/sops-install-secrets/sshkeys"
 	agessh "github.com/Mic92/ssh-to-age"
 
+	"github.com/facebookgo/symwalk"
```
Mic92, line 24: The repo is already archived. I would just copy this one function rather than depending on external code: https://github.com/facebookarchive/symwalk/blob/42004b9f322246749dd73ad71008b1f3160c0052/walk.go#L19

Mic92:
Can I get another force push to run drone here?

modules/sops/default.nix
```diff
@@ -133,6 +146,12 @@ in {
       '';
     };
 
+    log = mkOption {
+      type = types.listOf (types.enum [ "gpgImport" "changes" ]);
+      default = [ "gpgImport" "changes" ];
```
Mic92, line 151: Maybe we should rename this to `keyImport`? I assume most users will just use either gpg or age.

Mic92: I would rename `changes` to `serviceRestarts`.

dasJ: Problem is that it also prints changes when nothing is restarted, for example `adding secret: mysecret`

PR #114, opened 2021-09-10
Implement nested secrets

dasJ:


README.md
```diff
@@ -281,8 +281,13 @@ In our example we put the following content in it:
 example-key: example-value
 ```
 
-NOTE: At the moment we do not support nested data structures that
-sops support. This might change in the future. See also [Different file formats](#different-file-formats)
+Nesting the key results in the creation of directories.
+These directories will be owned by root:root and have permissions 0755.
```
Mic92, line 285: So this is what are currently doing for the root secret directory, and I think we should stick
with it for nesting:

```suggestion
These directories will be owned by root:keys and have permissions 0751.
```

pkgs/sops-install-secrets/main.go
```diff
@@ -307,12 +350,17 @@ func prepareSecretsDir(secretMountpoint string, linkName string, keysGid int) (*
 
 func writeSecrets(secretDir string, secrets []secret) error {
 	for _, secret := range secrets {
-		filepath := filepath.Join(secretDir, secret.Name)
-		if err := ioutil.WriteFile(filepath, []byte(secret.value), secret.mode); err != nil {
-			return fmt.Errorf("Cannot write %s: %w", filepath, err)
+		fp := filepath.Join(secretDir, secret.Name)
+		dir := filepath.Dir(fp)
+
+		if err := os.MkdirAll(dir, 0755); err != nil {
```
Mic92, line 356:
```suggestion
		if err := os.MkdirAll(dir, 0751); err != nil {
```

PR #118, opened 2021-09-28
also print imported age keys

Mic92:


pkgs/sops-install-secrets/main.go
```diff
@@ -534,15 +534,18 @@ func importAgeSSHKeys(keyPaths []string, ageFilePath string) error {
...
 			return fmt.Errorf("Cannot convert ssh key '%s': %w", p, err)
 		}
 		// Append it to the file
-		_, err = ageFile.WriteString(*bech32 + "\n")
+		_, err = ageFile.WriteString(*privKey + "\n")
 		if err != nil {
 			return fmt.Errorf("Cannot write key to age file: %w", err)
 		}
+
+		fmt.Printf("%s: Imported %s as Age key with fingerprint %s\n", path.Base(os.Args[0]), p, *pubKey)
```
Mic92, line 548: I also managed to encrypt secrets with shamir encryption by having the wrong .sops.yaml configuration.
So having this log message is actually quite helpful for debugging.

dasJ:
Not a huge fan of adding more output since #113 adds up to 3 lines of output. What do you think about only doing that on boot (→ when /run/secrets does not exist yet)?

Mic92:
> Not a huge fan of adding more output since #113 adds up to 3 lines of output. What do you think about only doing that on boot (→ when /run/secrets does not exist yet)?

This does not help really, users that migrate to sops, won't reboot their machines for testing.
If we don't have this than it  quite a pain to debug if users forgot to rotate keys after updating .sops.yaml
This is also relevant if users open up issues posting their yaml files and boot logs.

dasJ:
What about an environment variable like `SOPS_SILENT_KEY_IMPORT`? I could add another one to #113 like `SOPS_SILENT_CHANGE_DETECTION`, then everybody can fit the behaviour to their liking

dasJ:
I have implemented something similar but based on the manifest instead of the environment: https://github.com/Mic92/sops-nix/pull/113/commits/6473b947edad6d112ea7747f783e7f0219197d91

Mic92:
I think I want to only print imported keys when an error happens. This should please all parties and requires no option. Than I can also make error messages better when gnupgHome or age.keyFile is used.

PR #122, opened 2021-09-30
Swap order of age ssh keys and the key file

dasJ:
It makes more sense to import the key when we have one and ignore the
SSH keys instead of only importing the key when we have no SSH keys.
This is because we import all SSH keys by default in the module and
using a key file means the use has to explicitly unset the SSH keys.

### DIFF
--- a/pkgs/sops-install-secrets/main.go
+++ b/pkgs/sops-install-secrets/main.go
@@ -645,15 +645,15 @@ func installSecrets(args []string) error {
 		os.Setenv("GNUPGHOME", manifest.GnupgHome)
 	}
 
-	if len(manifest.AgeSshKeyPaths) != 0 {
+	if manifest.AgeKeyFile != "" {
+		os.Setenv("SOPS_AGE_KEY_FILE", manifest.AgeKeyFile)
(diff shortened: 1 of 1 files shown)


dasJ:
Note that this also makes the code actually compliant to the option docs I wrote:
```nix
{
  age = {
    keyFile = mkOption {
      description = ''
        Path to age key file used for sops decryption.
        Setting this to a non-null value causes the ssh keys to be ignored.
      '';
    };
    sshKeyPaths = mkOption {
      description = ''
        Paths to ssh keys added as age keys during sops description.
        This setting is ignored when the keyFile is set to a non-null value.
      '';
    };
  };
}
```

jlesquembre:
@dasJ does it mean that now, if I set the age key (e.g.: `sops.age.keyFile = "/var/lib/sops-nix/key.txt";`) I do **not** have to set `sops.age.sshKeyPaths = []`?

dasJ:
Yes it does. I wanted to ask you to test this, but the other two PRs (#123 and #121) should be merged first so the results are the best ;)

jlesquembre:
@dasJ all clear, thanks. I'm following the other 2 PRs, once merged, I'll test it again and report the result :-)

PR #127, opened 2021-10-19
Allow setting user passwords

### LINKED ISSUE #93: Does not work with `users.users.<name>.passwordFile`
If you attempt to do something like this:
```Nix
{
  sops.secrets.root-password.sopsFile = ./root-password.yaml;
  users.users.root.passwordFile = config.sops.secrets.root-password.path;
}
```
Where `root-password.yaml` looks like this:
```Yaml
root-password: $6$rounds=65536$00000000000000$00000000000000000000000000000000000000000000000000000000000000000000000000000000000000 
```

You end up with no password on boot for the user, looking at `/etc/shadow` shows `root:!:1::::::`. If you force reactivation to occur manually after booting by running `sudo /run/current-system/bin/switch-to-configuration switch` then `/etc/shadow` is updated and contains the correct password hash.

My guess is this is due to some annoying precedence issue, is there anyway around it?

dasJ:
Closes #93

### DIFF
--- a/README.md
+++ b/README.md
@@ -589,6 +589,25 @@ $ ls -la /var/lib/hass/secrets.yaml
 lrwxrwxrwx 1 root root 40 Jul 19 22:36 /var/lib/hass/secrets.yaml -> /run/secrets/home-assistant-secrets.yaml
 ```
 
+## Setting a user's password
+
+sops-nix has to run after users were created by NixOS.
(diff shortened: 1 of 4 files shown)


0x4A6F (approved):
Thanks. :heart: 

Tested and working with GnuPG only.

Mic92:
Works for me!

PR #136, opened 2021-11-12
Remount secrets read-only

dasJ:


pkgs/sops-install-secrets/main.go
```diff
@@ -330,6 +335,21 @@ func mountSecretFs(mountpoint string, keysGid int) error {
 	return nil
 }
 
+func remountSecretFsRo(mountpoint string) error {
```
Mic92, line 338: To reduce de-duplicate logic, I would propose this

```suggestion
func ensureSecretFs(mountpoint string, readonly bool) error {
}
```

This function would than use `Statfs` and to check if we need to mount it, or if we need to remount as read-only/read-write.

pkgs/sops-install-secrets/main.go
```diff
@@ -936,11 +957,15 @@ func installSecrets(args []string) error {
 				return fmt.Errorf("Cannot write key to age file: %w", err)
 			}
 		}
+		ageFile.Close()
```
Mic92, line 960: Can we move the importing the age file to its own function instead to make the `defer` happen earlier?

pkgs/sops-install-secrets/main.go
```diff
@@ -936,11 +957,15 @@ func installSecrets(args []string) error {
...
 	}
 
 	if err := decryptSecrets(manifest.Secrets); err != nil {
 		return err
 	}
+	if len(manifest.SSHKeyPaths) != 0 {
+		keyring.Remove()
```
Mic92, line 967: Maybe even the whole decryptSecrets part needs to be it's own function to reduce the scope of `keyring`.

Mic92: After checking `decryptSecrets`, we can just move setting up the keyring and importing the agefile into this function.

Mic92:
merge conflict unfortunately.

dasJ:
Seems to work:
```
root@helsinki-web03:/home/janne # mount | grep secrets
none on /run/secrets.d type ramfs (ro,nosuid,nodev,relatime,mode=751)
root@helsinki-web03:/home/janne # /run/current-system/bin/switch-to-configuration test
activating the configuration...
setting up /etc...
reloading user units for nix-deploy...
reloading user units for janne...
setting up tmpfiles
root@helsinki-web03:/home/janne # mount | grep secrets
none on /run/secrets.d type ramfs (ro,nosuid,nodev,relatime,mode=751)
```

pkgs/sops-install-secrets/main.go
```diff
@@ -308,24 +308,39 @@ func decryptSecrets(secrets []secret) error {
...
 	}
-
+	// Ensure the directory is properly owned
 	if err := os.Chown(mountpoint, 0, int(keysGid)); err != nil {
 		return fmt.Errorf("Cannot change owner/group of '%s' to 0/%d: %w", mountpoint, keysGid, err)
 	}
+	// Mount readonly if needed
+	if readonly {
+		if err := unix.Mount("", mountpoint, "ramfs", uintptr(mountFlags|unix.MS_RDONLY|unix.MS_REMOUNT), ""); err != nil {
+			return fmt.Errorf("Cannot mount as readonly: %s", err)
```
Mic92, line 340:
```suggestion
			return fmt.Errorf("Cannot mount as readonly (maybe blocked by a process): %s", err)
```

pkgs/sops-install-secrets/main.go
```diff
@@ -952,8 +977,11 @@ func installSecrets(args []string) error {
 	if err := handleModifications(isDry, manifest.Logging, manifest.SymlinkPath, *secretDir, manifest.Secrets); err != nil {
 		return fmt.Errorf("Cannot request units to restart: %w", err)
 	}
-	// No need to perform the actual symlinking
+	// No need to perform the actual symlinking, just remount as readonly
 	if isDry {
+		if err := ensureSecretFs(manifest.SecretsMountPoint, true, keysGid); err != nil {
```
Mic92, line 982: I still have not figured out why this sometimes fail on some hosts. Can we keep it behind a feature flag until the issue is resolved?

dasJ:
Rebased and made the feature disabled by default. Does this fit you better?

Mic92:
I had issues with it back than and the benefit of having it read-only seems not so big given the complexity it adds (also w.r.t. to concurrent nixos updates).

PR #146, opened 2021-11-28
Remove restart logic from README and test

dasJ:
The required code in nixpkgs was reverted so we should not advertise a
feature that does not work. We can revert this commit if the feature is
re-merged into 22.05 with the proper version in it.

### DIFF
--- a/pkgs/sops-install-secrets/nixos-test.nix
+++ b/pkgs/sops-install-secrets/nixos-test.nix
@@ -205,108 +205,108 @@
  };
 
 } // pkgs.lib.optionalAttrs (pkgs.lib.versionAtLeast (pkgs.lib.versions.majorMinor pkgs.lib.version) "21.11") {
+  # This feature got reverted in nixpkgs...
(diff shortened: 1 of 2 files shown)


pkgs/sops-install-secrets/nixos-test.nix
```diff
@@ -203,110 +203,4 @@
    inherit pkgs;
    inherit (pkgs) system;
  };
-
-} // pkgs.lib.optionalAttrs (pkgs.lib.versionAtLeast (pkgs.lib.versions.majorMinor pkgs.lib.version) "21.11") {
```
Mic92, line 207: To avoid merge conflicts, we can keep the test code here, with a comment that we currently not officially support this feature.

Mic92:
> The required code in nixpkgs was reverted so we should not advertise a feature that does not work. We can revert this commit if the feature is re-merged into 22.05 with the proper version in it.

Does it not work for simple restarts even?

PR #155, opened 2022-01-14
allow to disable sops

Mic92:
code is completely untested, not sure it even evals.

### DIFF
--- a/modules/sops/default.nix
+++ b/modules/sops/default.nix
@@ -14,6 +14,13 @@ let
       sopsFileHash = mkOptionDefault (optionalString cfg.validateSopsFiles "${builtins.hashFile "sha256" config.sopsFile}");
     };
     options = {
+      enable = mkOption {
+        type = types.bool;
+        default = true;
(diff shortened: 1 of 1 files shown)


dasJ:
Huh why? I could also just do `sops.secrets = lib.mkOverride 0 {} `

Mic92:
> Huh why? I could also just do `sops.secrets = lib.mkOverride 0 {} `

I want to test for `sops.enable` in my modules and make them conditionally fall back to some other value if no secrets are available for this host.

modules/sops/default.nix
```diff
@@ -14,6 +14,13 @@ let
       sopsFileHash = mkOptionDefault (optionalString cfg.validateSopsFiles "${builtins.hashFile "sha256" config.sopsFile}");
     };
     options = {
+      enable = mkOption {
```
dasJ, line 17:
```suggestion
      enable = mkEnableOption "sops secrets" // { default = true; };
```

PR #168, opened 2022-03-29
Implement home-manager support

### LINKED ISSUE #62: Does `nix-sops` support using it as a module in `home-manager` configuration?
If it is currently supported, please offer documentation about how to set it up.

If it is currently *not* supported, please consider supporting it.

### LINKED ISSUE #163: nix-darwin support
Tangentially related to #62, it would be useful to be able to ship secrets on nix-darwin systems.

dasJ:
Closes #62
Closes #163

### DIFF
--- a/README.md
+++ b/README.md
@@ -741,7 +741,7 @@ This module provides a subset of features provided by the system-wide sops-nix s
 
 Instead of running as an activation script, sops-nix runs as a systemd user service called `sops-nix.service`.
 And instead of decrypting to `/run/secrets`, the secrets are decrypted to `$XDG_RUNTIME_DIR/secrets`.
-**Since the secrets are decryted there, it's highly recommended to use a tmpfs for `$XDG_RUNTIME_DIR` if your distribution does not do that.**
+**Since the secrets are decrypted there, it's highly recommended to use a tmpfs for `$XDG_RUNTIME_DIR` to avoid storing secrets in plain text on persistent storage. Linux distributions using systemd-logind do that out-of-the-box.**
 
 Depending on whether you use home-manager system-wide or using a home.nix, you have to import it in a different way.
 This example show the `channel` approach from the usage example above for simplicity, but all other methods work as well.
@@ -768,11 +768,11 @@ The actual sops configuration is in the `sops` namespace in your home.nix (or in
 ```nix
 {
   sops = {
-    age.keyFile = "/home/user/.age-key.txt" ]; # must have no password!
+    age.keyFile = "/home/user/.age-key.txt"; # must have no password!
     # It's alos possible to use a ssh key, but only when it has no password:
     #age.sshKeyPaths = [ "/home/user/path-to-ssh-key" ];
     sops.secrets.test = {
-      sopsFile = ./secrets.yml.enc;
+      # sopsFile = ./secrets.yml.enc; # optionally define per-secret files
       path = "%r/test.txt"; # %r gets replaced with your $XDG_RUNTIME_DIR, use %% to specify a '%' sign
     };
   };
--- a/README.md
+++ b/README.md
@@ -771,6 +771,7 @@ The actual sops configuration is in the `sops` namespace in your home.nix (or in
     age.keyFile = "/home/user/.age-key.txt"; # must have no password!
     # It's alos possible to use a ssh key, but only when it has no password:
     #age.sshKeyPaths = [ "/home/user/path-to-ssh-key" ];
+    defaultSopsFile = ./secrets.yaml;
     sops.secrets.test = {
       # sopsFile = ./secrets.yml.enc; # optionally define per-secret files
       path = "%r/test.txt"; # %r gets replaced with your $XDG_RUNTIME_DIR, use %% to specify a '%' sign
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ environment variables that can be passed to sops.
 - Compatible with all NixOS deployment frameworks: [NixOps](https://github.com/NixOS/nixops), nixos-rebuild, [krops](https://github.com/krebs/krops/), [morph](https://github.com/DBCDK/morph), [nixus](https://github.com/Infinisil/nixus), etc.
 - Version-control friendly: Since all files are encrypted they can be directly committed to version control without worry. Diffs of the secrets are readable, and [can be shown in cleartext](https://github.com/mozilla/sops#showing-diffs-in-cleartext-in-git).
 - CI friendly: Since sops files can be added to the Nix store without leaking secrets, a machine definition can be built as a whole from a repository, without needing to rely on external secrets or services.
+- Home-manager friendly: Provides a home-manager module
 - Works well in teams: sops-nix comes with `nix-shell` hooks that allows multiple people to quickly import all GPG keys.
   The cryptography used in sops is designed to be scalable: Secrets are only encrypted once with a master key
   instead of encrypted per machine/developer key.
@@ -34,7 +35,7 @@ There is a `configuration.nix` example in the [deployment step](#deploy-example)
 
 ## Supported encryption methods
 
-sops-nix supports two basic ways of encryption, GPG and `age`. 
+sops-nix supports two basic ways of encryption, GPG and `age`.
 
 GPG is based on [GnuPG](https://gnupg.org/) and encrypts against GPG public keys. Private GPG keys may
 be used to decrypt the secrets on the target machine. The tool [`ssh-to-pgp`](https://github.com/Mic92/ssh-to-pgp) can
@@ -729,6 +730,38 @@ This is how it can be included in your `configuration.nix`:
   sops.secrets.krb5-keytab = {
     format = "binary";
     sopsFile = ./krb5.keytab;
+  };
+}
+```
+
+## Use with home manager
+
+sops-nix also provides a home-manager module.
+This module provides a subset of features provided by the system-wide sops-nix since features like the creation of the ramfs and changing the owner of the secrets are not available for non-root users.
+
+Instead of running as an activation script, sops-nix runs as a systemd user service called `sops-nix.service`.
+And instead of decrypting to `/run/secrets`, the secrets are decrypted to `$XDG_RUNTIME_DIR/secrets`.
+
+Usage example:
+```nix
+{
+  # NixOS home-manager configuration
+  home-manager.sharedModules = [
+    /path/to/sops-nix/modules/home-manager/sops.nix
+  ];
+
+  # Configuration via home.nix
+  imports = [
+    /path/to/sops-nix/modules/home-manager/sops.nix
+  ];
+
+  # Configuration of secrets
+  sops = {
+    age.sshKeyPaths = [ "/home/user/path-to-ssh-key" ]; # must have no password!
+    sops.secrets.test = {
+      sopsFile = ./secrets.yml.enc;
+      path = "%r/test.txt"; # %r gets replaced with your $XDG_RUNTIME_DIR
+    };
   };
 }
 ```
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ If you use experimental nix flakes support:
   inputs.sops-nix.url = github:Mic92/sops-nix;
   # optional, not necessary for the module
   #inputs.sops-nix.inputs.nixpkgs.follows = "nixpkgs";
-  
+
   outputs = { self, nixpkgs, sops-nix }: {
     # change `yourhostname` to your actual hostname
     nixosConfigurations.yourhostname = nixpkgs.lib.nixosSystem {
@@ -714,7 +714,7 @@ $ head krb5.keytab
                 "mac": "ENC[AES256_GCM,data:ISjUzaw/5mNiwypmUrOk2DAZnlkbnhURHmTTYA3705NmRsSyUh1PyQvCuwglmaHscwl4GrsnIz4rglvwx1zYa+UUwanR0+VeBqntHwzSNiWhh7qMAQwdUXmdCNiOyeGy6jcSDsXUeQmyIWH6yibr7hhzoQFkZEB7Wbvcw6Sossk=,iv:UilxNvfHN6WkEvfY8ZIJCWijSSpLk7fqSCWh6n8+7lk=,tag:HUTgyL01qfVTCNWCTBfqXw==,type:str]",
                 "pgp": [
                         {
-                        
+
 ```
 
 It can be decrypted again like this:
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -747,6 +748,38 @@ This is how it can be included in your `configuration.nix`:
 }
 ```
 
+## Use with home manager
+
+sops-nix also provides a home-manager module.
+This module provides a subset of features provided by the system-wide sops-nix since features like the creation of the ramfs and changing the owner of the secrets are not available for non-root users.
+
+Instead of running as an activation script, sops-nix runs as a systemd user service called `sops-nix.service`.
+And instead of decrypting to `/run/secrets`, the secrets are decrypted to `$XDG_RUNTIME_DIR/secrets`.
```
Mic92, line 757: `$XDG_RUNTIME_DIR` can be not set on non-nixos. We should recommend to have it point it to a tmpfs so that secrets are not persisted stored in plain text or otherwise live with the consequences...

Mic92, line 757: `$XDG_RUNTIME_DIR` can be not set on non-nixos. We should recommend to have it point it to a tmpfs so that secrets are not persisted stored in plain text or otherwise live with the consequences...

README.md
```diff
@@ -747,6 +748,38 @@ This is how it can be included in your `configuration.nix`:
...
+
+Usage example:
+```nix
+{
+  # NixOS home-manager configuration
+  home-manager.sharedModules = [
+    /path/to/sops-nix/modules/home-manager/sops.nix
```
Mic92, line 764: Maybe fetchTarball could work here? I would also expose the module via flake.nix.

Mic92, line 764: Maybe fetchTarball could work here? I would also expose the module via flake.nix.

README.md
```diff
@@ -747,6 +748,38 @@ This is how it can be included in your `configuration.nix`:
...
+  ];
+
+  # Configuration via home.nix
+  imports = [
+    /path/to/sops-nix/modules/home-manager/sops.nix
+  ];
```
Mic92, line 770: It's confusing that home-manager and nixos configuration are in one code block. I would also keep the `home-manager` configuration below separate from importing sops-nix.

Mic92, line 770: It's confusing that home-manager and nixos configuration are in one code block. I would also keep the `home-manager` configuration below separate from importing sops-nix.

README.md
```diff
@@ -747,6 +748,38 @@ This is how it can be included in your `configuration.nix`:
...
+
+  # Configuration of secrets
+  sops = {
+    age.sshKeyPaths = [ "/home/user/path-to-ssh-key" ]; # must have no password!
+    sops.secrets.test = {
+      sopsFile = ./secrets.yml.enc;
+      path = "%r/test.txt"; # %r gets replaced with your $XDG_RUNTIME_DIR
```
Mic92, line 777: In that case we should also replace `%%` with `%` without expanding it for escaping purpose.

```suggestion
      path = "%r/test.txt"; # %r gets replaced with your $XDG_RUNTIME_DIR
```

Mic92, line 777: In that case we should also replace `%%` with `%` without expanding it for escaping purpose.

```suggestion
      path = "%r/test.txt"; # %r gets replaced with your $XDG_RUNTIME_DIR
```

README.md
```diff
@@ -747,6 +748,38 @@ This is how it can be included in your `configuration.nix`:
...
+  ];
+
+  # Configuration via home.nix
+  imports = [
+    /path/to/sops-nix/modules/home-manager/sops.nix
+  ];
+
+  # Configuration of secrets
+  sops = {
+    age.sshKeyPaths = [ "/home/user/path-to-ssh-key" ]; # must have no password!
```
Mic92, line 774: I am inclined to make normal age keys the default here as a best practice. Most of the time you want to have ssh keys encrypted because of their actual use.

Mic92, line 774: I am inclined to make normal age keys the default here as a best practice. Most of the time you want to have ssh keys encrypted because of their actual use.

pkgs/sops-install-secrets/main.go
```diff
@@ -870,6 +884,21 @@ func installSecrets(args []string) error {
 		return err
 	}
 
+	if manifest.UserMode {
+		rundir, ok := os.LookupEnv("XDG_RUNTIME_DIR")
+		if !ok {
+			rundir = fmt.Sprintf("/run/user/%d", os.Getuid())
```
Mic92, line 890: if `XDG_RUNTIME_DIR` is not set, it's likely that there is no `/run/user/<uid>`. I would rather return an error here rather than writing to some random directory.

Mic92, line 890: if `XDG_RUNTIME_DIR` is not set, it's likely that there is no `/run/user/<uid>`. I would rather return an error here rather than writing to some random directory.

modules/home-manager/sops.nix
```diff
@@ -0,0 +1,226 @@
...
+        message = "Cannot find path '${secret.sopsFile}' set in sops.secrets.${lib.strings.escapeNixIdentifier name}.sopsFile";
+      } {
+        assertion =
+          builtins.isPath secret.sopsFile ||
+          (builtins.isString secret.sopsFile && lib.hasPrefix builtins.storeDir secret.sopsFile);
+        message = "'${secret.sopsFile}' is not in the Nix store. Either add it to the Nix store or set sops.validateSopsFiles to false";
+      }]) cfg.secrets)
+    );
+
+    systemd.user.services.sops-nix = {
```
Mic92, line 205: We should have documentation on how to have services depending on secrets waiting for this service.

Mic92, line 205: We should have documentation on how to have services depending on secrets waiting for this service.

Kranzes:
The decryption of the keys should be part of home `home.activation`.

Mic92:
> The decryption of the keys should be part of `home.activation`.

`home.activation` is not run on boot, which means we also need a service to decrypt secrets to tmpfs.
Because of this we already need to have dependencies between services, I don't see the value added by also having it in activation phase. Does this make sense?

Kranzes:
I did some further testing yesterday and it did start on ("boot") login to the user.

Mic92:
> I did some further testing yesterday and it did start on ("boot") login to the user.

Ok. Than I need to investigate this.

Kranzes:
I recommend you to test it as well. also I did some work on agenix home-manager module yesterday, https://github.com/ryantm/agenix/pull/109, you might want to take a look. Also I mentioned all the challenges I encountered on the agenix matrix channel if you want to go and read them.

NobbZ:
@Kranzes 

> I did some further testing yesterday and it did start on ("boot") login to the user.

AFAIU this is only true when using `home-manager` as a module for nixos, not when used standalone. As activation is implemented as a user service on nixos then.

Therefore I think using services which are `Before` the user activation if and only if used as a service, but run standalone on standalone.

I have no clue how to solve this on darwin though. I do not have any experience with that system, I just know that there is no equivalent to systemd user scervices.

mtoohey31:
> I have no clue how to solve this on darwin though. I do not have any experience with that system, I just know that there is no equivalent to systemd user scervices.

Darwin has launchd, and it looks like home-manager has [a module](https://github.com/nix-community/home-manager/tree/master/modules/launchd) for it, so that might be a good solution.

otavio:
Looking forward to use this in my nix-config :-D

dasJ:
@mtoohey31 could you help with the launchd logic? I don't really have the infrastructure to test anything in that regard

dasJ:
I think I have pretty much all review comments fixed now.
Still missing is launchd support

dasJ:
Added an untested launchd module, please test

mtoohey31:
> Added an untested launchd module, please test

Sorry for the slow response, I'd be happy to help with testing but I'm out of time tonight. I'll try to get to it soon.

README.md
```diff
@@ -733,6 +734,58 @@ This is how it can be included in your `configuration.nix`:
 }
 ```
 
+## Use with home manager
```
Mic92, line 737: Reminder to myself to expose this in the sops flake as well.

modules/home-manager/sops.nix
```diff
@@ -0,0 +1,240 @@
...
+          Paths to ssh keys added as age keys during sops description.
+        '';
+      };
+    };
+
+    gnupg = {
+      home = lib.mkOption {
+        type = lib.types.nullOr lib.types.str;
+        default = null;
+        example = "/root/.gnupg";
```
Mic92, line 180:
```suggestion
        example = "/home/someuser/.gnupg";
```

modules/home-manager/sops.nix
```diff
@@ -0,0 +1,240 @@
...
+      type = lib.types.listOf (lib.types.enum [ "keyImport" "secretChanges" ]);
+      default = [ "keyImport" "secretChanges" ];
+      description = "What to log";
+    };
+
+    age = {
+      keyFile = lib.mkOption {
+        type = lib.types.nullOr lib.types.path;
+        default = null;
+        example = "/var/lib/sops-nix/key.txt";
```
Mic92, line 150: I would point in the example to somewhere in `/home` maybe.

README.md
```diff
@@ -733,6 +734,58 @@ This is how it can be included in your `configuration.nix`:
...
+
+sops-nix also provides a home-manager module.
+This module provides a subset of features provided by the system-wide sops-nix since features like the creation of the ramfs and changing the owner of the secrets are not available for non-root users.
+
+Instead of running as an activation script, sops-nix runs as a systemd user service called `sops-nix.service`.
+And instead of decrypting to `/run/secrets`, the secrets are decrypted to `$XDG_RUNTIME_DIR/secrets`.
+**Since the secrets are decryted there, it's highly recommended to use a tmpfs for `$XDG_RUNTIME_DIR` if your distribution does not do that.**
```
Mic92, line 744:
```suggestion
**Since the secrets are decrypted there, it's highly recommended to use a tmpfs for `$XDG_RUNTIME_DIR` to avoid storing secrets in plain text on persistent storage. Linux distributions using systemd-logind do that out-of-the-box.**
```

README.md
```diff
@@ -733,6 +734,58 @@ This is how it can be included in your `configuration.nix`:
...
+    <sops-nix/modules/home-manager/sops.nix>
+  ];
+}
+```
+
+The actual sops configuration is in the `sops` namespace in your home.nix (or in the `home-manager.users.<name>` namespace when using home-manager system-wide):
+```nix
+{
+  sops = {
+    age.keyFile = "/home/user/.age-key.txt" ]; # must have no password!
```
pogobanane, line 771:
```suggestion
    age.keyFile = "/home/user/.age-key.txt"; # must have no password!
```

README.md
```diff
@@ -733,6 +734,58 @@ This is how it can be included in your `configuration.nix`:
...
+    # It's alos possible to use a ssh key, but only when it has no password:
+    #age.sshKeyPaths = [ "/home/user/path-to-ssh-key" ];
+    sops.secrets.test = {
+      sopsFile = ./secrets.yml.enc;
```
pogobanane, line 775:
```suggestion
    defaultSopsFile = ./secrets.yml.enc;
    sops.secrets.test = {
      # sopsFile = ./secrets.yml.enc; # optionally define per-secret files
```

pogobanane (commented):
I have to `systemctl --user restart sops-nix` manually after every change. Can this be automated?

pkgs/sops-install-secrets/main.go
```diff
@@ -870,6 +897,21 @@ func installSecrets(args []string) error {
 		return err
 	}
 
+	if manifest.UserMode {
+		rundir, ok := os.LookupEnv("XDG_RUNTIME_DIR")
+		if !ok {
+			return fmt.Errorf("$XDG_RUNTIME_DIR is not set!")
```
pogobanane, line 903: I'm building my home-manager config with root as part of my nixos-config. `Nixos-rebuild switch` fails for me with:
```
error: builder for '/nix/store/bhscqh6gvv32cda919mkh2sa7yaa9chi-manifest.json.drv' failed with exit code 1;
       last 1 log lines:
       > /nix/store/ppl4vbv67c420zk9r87izdsfwms5q2ja-sops-install-secrets-0.0.1/bin/sops-install-secrets: $XDG_RUNTIME_DIR is not set!
       For full logs, run 'nix log /nix/store/bhscqh6gvv32cda919mkh2sa7yaa9chi-manifest.json.drv'.
error: 1 dependencies of derivation '/nix/store/a93kndskwslvnaaymqqk4svjhj7s33gg-sops-nix-user.drv' failed to build
```
I suspect this happens during the check phase. It may be sufficient to just ignore `$XDG_RUNTIME_DIR` during `--check-mode`.

pogobanane, line 903: this fixes my home-manager issue:
```suggestion
		if opts.checkMode == Off && !ok {
		        // Sometimes builds run check-mode outside the users environment. Thus
		        // this env var may not be set at check time, but it must be at runtime.
			return fmt.Errorf("$XDG_RUNTIME_DIR is not set!")
```

README.md
```diff
@@ -733,6 +734,58 @@ This is how it can be included in your `configuration.nix`:
...
+```
+
+The actual sops configuration is in the `sops` namespace in your home.nix (or in the `home-manager.users.<name>` namespace when using home-manager system-wide):
+```nix
+{
+  sops = {
+    age.keyFile = "/home/user/.age-key.txt"; # must have no password!
+    # It's alos possible to use a ssh key, but only when it has no password:
+    #age.sshKeyPaths = [ "/home/user/path-to-ssh-key" ];
+    sops.secrets.test = {
```
pogobanane, line 774:
```suggestion
    #age.sshKeyPaths = [ "/home/user/path-to-ssh-key" ];
    defaultSopsFile = ./secrets.yaml;
    sops.secrets.test = {
```

pogobanane: this is required, but got lost somehow

pogobanane (approved):
maybe wait for @mtoohey31 s feedback for macos, but you have my approval for nixos.

mtoohey31 (changes requested):
Sorry that it's taken me a bit to get to this. In addition to the other review comment, `sops-install-secrets` requires some work before I can fully test this.

`default.nix` needs to be updated so `sops-install-secrets` is available for darwin, and `pkgs/sops-install-secrets/default.nix` needs to have `platforms.darwin` added to the supported platforms.

Getting `sops-install-secrets` to actually build on MacOS will require some work too. We need to remove the `// +build linux` constraints from `main.go` and `main_test.go`, and after that there are still some issues with uses of the `unix` package:

```
# github.com/Mic92/sops-nix/pkgs/sops-install-secrets
./main.go:119:54: undefined: unix.O_PATH
./main.go:133:44: undefined: unix.AT_EMPTY_PATH
./main.go:328:23: too many arguments in call to unix.Mount
./main.go:328:53: undefined: unix.MS_NODEV
./main.go:328:67: undefined: unix.MS_NOSUID
```

I believe we'll need Linux and MacOS-specific versions of the `secureSymlinkChown` and `mountSecretFs` functions. That probably shouldn't be too much work, but I can't claim to understand all the security details of those functions. Does anyone else who's more familiar with the code have access to a Mac?

pogobanane:
i started a darwin home-manager implementation based on this PR here: https://github.com/pogobanane/sops-nix/tree/feat/home-manager-darwin 
- which compiles
- whose git history needs to be cleaned up
- which is completely untested
- which needs error handling
- whose workarounds/TODOs need work

Also there is https://github.com/4825764518/sops-nix/commits/darwin which implements a darwinModule (nixosModule for MacOs, _not_ home-manager module).

mtoohey31:
Sorry for the slow reply, your fork works for me! I made two small changes [here](https://github.com/mtoohey31/sops-nix/tree/feat/home-manager-darwin), and with those changes I was able to set up a secret for my github cli credentials on [this branch](https://github.com/mtoohey31/infra/tree/feat/home-manager-sops) of my personal configuration. The secret is being set up properly on reboot too.

Aside from that, I haven't done any extensive testing. I'm not familiar with all of `sops-nix`'s features, so feel free to add on if I've missed anything, but here's some of the cases I figure I should still check:

- [ ] Different formats
- [ ] Different modes
- [ ] Non-default `sopsFile` values

Mic92:
Does nix-darwin supports some sort of unitests? I am not a regular macos user and I fear I might break it accidentally.

mtoohey31:
> Does nix-darwin supports some sort of unitests? I am not a regular macos user and I fear I might break it accidentally.

Tests would probably be a good idea. Both nix-darwin and home-manager have tests included in their main repositories. nix-darwin's are kind of ad-hoc and not really extensible as far as I can tell, but home-manager uses [nmt](https://gitlab.com/rycee/nmt); it might be helpful here.

mtoohey31:
I've finished some more testing; it looks like all the things I had listed on the checklist above are working properly. I would recommend setting an output path for the launchd agent's logs though, as I have in d6eb1b8e4d95285a8c1843920dbb7554b22f949a.

anilanar:
I'm waiting for it to be exposed in flakes to test it.

dasJ:
Feel free to open a PR against this branch if you need this feature.

anilanar:
@dasJ Will do so once I manage to make it work.

Right now, systemd service fails to start. One thing I noticed is that there's a missing space somewhere:

```
> cat /nix/store/jyk79akz7j478pr2f4kfg9pid0b88vs5-sops-nix-user
#!/nix/store/iffl6dlplhv22i2xy7n1w51a5r631kmi-bash-5.1-p16/bin/bash
export SOPS_GPG_EXEC=/nix/store/2ypi8rh1nm1v2w2pafva4rk1vigsjjvj-gnupg-2.3.4/bin/gpg/nix/store/7zi58cdmn8csqi59sp91f1f3bwi9lwcb-sops-install-secrets-0.0.1/bin/sops-install-secrets -ignore-passwd '/nix/store/1nf3f2pssq0nj5p4frcx0yhrplbpqxjq-manifest.json'
```

There's no space between `gpg` and `sops-install-secrets` paths.

otavio:
Is anything still pending? I look forward to this feature.

Mic92:
Related https://github.com/jordanisaacs/homeage

anilanar:
I've been using this branch with gpg for a while now, with some fixes applied at https://github.com/Mic92/sops-nix/compare/master...anilanar:sops-nix:feat/home-manager-flake.

Maybe merge and wait for user feedback? I don't see any way to make progress but wait for people to test this if it works.

Mic92:
@anilanar sure, just open a new PR and I will review it.

anilanar:
Let's merge this? One minor note though: this PR doesn't fix #163 unlike what PR description says, if I'm not mistaken.

G3zz:
Are there any blockers on this now?

pogobanane:
i merged the home-manager-darwin version with this branch in #261. @mtoohey31 could you do another quick test if it still works on darwin?

Mic92:
Closing this one. since the @pogobanane one is now more advanced.

mtoohey31:
> @mtoohey31 could you do another quick test if it still works on darwin?

Sorry that I never replied to this. I see here: https://github.com/Mic92/sops-nix/pull/261#issuecomment-1412683822 that you were able to test things on darwin, so I take it there's no need for me to do any testing at this point?

PR #171, opened 2022-04-12
doc: fix path towards the nobody user

teto:


Mic92:
Thanks!

teto:
thank you for the superb documentation and the module. I would have spent so much more time without your very nice readme.

PR #195, opened 2022-06-23
README: remove mention of decrypting SSH private key

winterqt:
The instructions provided point `ssh-to-age` at a public key, which won't (ever?) be encrypted. This removes the SSH private key decryption mention, to not confuse readers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,20 +195,6 @@ $ nix-shell -p ssh-to-age --run "ssh-to-age < ~/.ssh/id_ed25519.pub"
 $ nix-shell -p ssh-to-age --run "ssh-add -L | ssh-to-age"
 ```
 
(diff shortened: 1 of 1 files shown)


winterqt:
Now that I think about it, the docs don't specify having to leave the (unprotected) age private key anywhere for SOPS to see it after converting (or having to convert the private key at all, for that matter).

winterqt:
> Now that I think about it, the docs don't specify having to leave the (unprotected) age private key anywhere for SOPS to see it after converting (or having to convert the private key at all, for that matter).

@Mic92 I still think is something that should be addressed, probably by expanding that portion of the documentation to include converting the private key. Would you be open to reviewing a PR that made this change?

PR #208, opened 2022-07-12
Add validationPackage option for cross-compilation

dramforever:
Separate out `sops-install-secrets` into a 'runtime version' and 'check version', so that even if the builder cannot run target code it still works.

Works a bit like the sshd module in nixpkgs:

https://github.com/NixOS/nixpkgs/blob/d632b9657c7dd5c994e5cdf0bc71cb016f63efa1/nixos/modules/services/networking/ssh/sshd.nix#L7-L13

### DIFF
--- a/modules/sops/default.nix
+++ b/modules/sops/default.nix
@@ -6,6 +6,7 @@ let
   cfg = config.sops;
   users = config.users.users;
   sops-install-secrets = cfg.package;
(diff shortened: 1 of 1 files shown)


dasJ (approved):

ncfavier:
Please remember to add `defaultText` to options with non-trivial defaults (https://github.com/Mic92/sops-nix/pull/211)

PR #211, opened 2022-07-21
module: add defaultText to validationPackage

ncfavier:
Makes flake-info succeed


dramforever (approved):
This looks great. Thanks for the docs fix!

PR #257, opened 2023-01-05
Adding support for generating a Curve25119 based GPG key with sops-init-gpg-key

kuetemeier:
This add the option `--keytype` to `sops-init-gpg-key`. If it is missing, the default behaviour is kept and it generates a RSA GPG key - like before. If it is added like `--keytype Curve25119` (see README commit) it will generate a Curve25119 based key. Everything else works as before.

### DIFF
--- a/pkgs/sops-init-gpg-key/sops-init-gpg-key
+++ b/pkgs/sops-init-gpg-key/sops-init-gpg-key
@@ -3,7 +3,7 @@
 set -o errexit -o pipefail -o noclobber -o nounset
 
 OPTIONS=h
-LONGOPTS=help,gpghome:,hostname:
+LONGOPTS=help,gpghome:,hostname:,keytype:
 
 ! PARSED=$(getopt --options=$OPTIONS --longoptions=$LONGOPTS --name "$0" -- "$@")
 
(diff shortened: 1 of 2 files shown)


kuetemeier:
Because I have a complet setup with YubiKeys and GPG. I like Curve25119 more than RSA (feels safer to me). I used this with a lot of servers in 2022. Just found the time to make a pull request ;-)

pkgs/sops-init-gpg-key/sops-init-gpg-key
```diff
@@ -56,10 +64,28 @@ fi
...
+if [[ "$KEYTYPE" == "Curve25519" ]]; then
+  cat >> "$GNUPGHOME/key-template" <<EOF
+Key-Type: eddsa
+Key-Curve: Ed25519
+Key-Usage: sign
+Subkey-Type: ecdh
+Subkey-Curve: Curve25519
+Subkey-Usage: encrypt
+EOF
+else
```
Mic92, line 81:
```suggestion
elif [[ "$KEYTYPE" == "Curve25519" ]]; then
```

pkgs/sops-init-gpg-key/sops-init-gpg-key
```diff
@@ -56,10 +64,28 @@ fi
...
+Subkey-Type: ecdh
+Subkey-Curve: Curve25519
+Subkey-Usage: encrypt
+EOF
+else 
+  cat >> "$GNUPGHOME/key-template" <<EOF
 Key-Type: 1
 Key-Length: 2048
+EOF
+fi
```
Mic92, line 86:
```suggestion
else
  echo "unknown keytype '$KEYTYPE'"
  exit 1
fi
```

PR #262, opened 2023-01-16
format type: add dotenv

lucasew:
Sops in version [3.2.0](https://gist.github.com/ageis/f5595e59b1cddb1513d1b425a323db04) introduced support of the dotenv format. It's working pretty well here, but sops-nix doesn't allow it yet.

This pull request just allows sops-nix users to integrate dotenv based secrets.


Mic92 (approved):
like the idea; it would be nice to also have this format tested in CI.

lucasew:
Houston, we have a problem

The go part of sops-nix doesn't handle this. At first I though that the project used pure sops instead of a wrapper so the hole is deeper than I though.

lucasew:
How hard would be to port this project to be a sops + ssh-to-age wrapper?

Mic92:
Shouldn't be that hard to also add dotenv here for de-serialization: https://github.com/Mic92/sops-nix/pull/262/files#diff-5ebcd22450b1d2333c6ae4ac07ebca561b4f3ac2b6ea9e38d4c0049503775d49R277

lucasew:
It's more complex than this.

I don't know exactly what I should do, but it's definitely not only this.

Mic92:
Well, do we need to upgrade our sops library somehow so it understands the new format?

Mic92:
Also we are at already: v3.7.3 https://github.com/Mic92/sops-nix/blob/e18eefd2b133a58309475298052c341c08470717/go.mod#L23

lucasew:
> Well, do we need to upgrade our sops library somehow so it understands the new format?

No

The lib already has all the stuff we need

Mic92:
Thanks!

Issue #295, opened 2023-03-10
keygroups in .sops.yaml examples


vdbe:
Every example of `.sops.yaml` in the README creates 2 key groups (1 for pgp, 1 for age).

This sets the shamir threshold to 2, I think this is a mistake in the readme as the [more complex example](https://github.com/Mic92/dotfiles/blob/master/nixos/.sops.yaml) does not do this.

changing
```yaml
# This example uses YAML anchors which allows reuse of multiple keys 
# without having to repeat yourself.
# Also see https://github.com/Mic92/dotfiles/blob/master/nixos/.sops.yaml
# for a more complex example.
keys:
  - &admin_alice 2504791468b153b8a3963cc97ba53d1919c5dfd4
  - &admin_bob age12zlz6lvcdk6eqaewfylg35w0syh58sm7gh53q5vvn7hd7c6nngyseftjxl
creation_rules:
  - path_regex: secrets/[^/]+\.yaml$
    key_groups:
    - pgp:
      - *admin_alice
    - age:
      - *admin_bob
```
to
```yaml
# This example uses YAML anchors which allows reuse of multiple keys 
# without having to repeat yourself.
# Also see https://github.com/Mic92/dotfiles/blob/master/nixos/.sops.yaml
# for a more complex example.
keys:
  - &admin_alice 2504791468b153b8a3963cc97ba53d1919c5dfd4
  - &admin_bob age12zlz6lvcdk6eqaewfylg35w0syh58sm7gh53q5vvn7hd7c6nngyseftjxl
creation_rules:
  - path_regex: secrets/[^/]+\.yaml$
    key_groups:
    - pgp:
      - *admin_alice
      age:
      - *admin_bob
```
would fix this and align it with the more complex example.

PR #296, opened 2023-03-10
fix(readme): keygroups in .sops.yaml examples

vdbe:
Fixes: #295


Mic92:
Thanks. That's indeed a crucial difference...

PR #299, opened 2023-03-13
Add `sops.templates`

Ninlives:
# Motivation
This pull request adds `sops.templates` option, which allows users to encrypt only the credential part of a file, to improve the readability and maintainability of code.

# Usage Example

```nix
{ ... }: {
  sops.secrets.v2ray-user = {};
  sops.secrets.v2ray-pass = {};

  services.v2ray = {
    enable = true;
    configFile = config.sops.templates.v2ray.path; # The path of the rendered secret file
  };

  sops.templates.v2ray.content = builtins.toJSON {
    inbounds = [{
      port = 1080;
      listen = "127.0.0.1";
      protocol = "http";
      settings = {
        accounts = [{
          # These placeholders will be replaced by the real credentials during activation.
          user = config.sops.placeholder.v2ray-user;
          pass = config.sops.placeholder.v2ray-pass;
        }];
      };
    }];
    outbounds = [{
      protocol = "freedom";
    }];
  };
}
```


modules/sops/templates/subs.py
```diff
@@ -0,0 +1,17 @@
...
+
+with open(subst) as f:
+    subst_pairs = f.read().splitlines()
+
+for pair in subst_pairs:
+    placeholder, path = pair.split()
+    with open(path) as f:
+        content = content.replace(placeholder, f.read())
+
+print(content)
```
Mic92, line 17: Would be nice to wrap in this in a `def main() -> None:` function so that it's type checked by mypy.

modules/sops/templates/default.nix
```diff
@@ -0,0 +1,108 @@
...
+          ++ optional (secretsForUsers != { }) "setupSecretsForUsers") ''
+            echo Setting up sops templates...
+            ${concatMapStringsSep "\n" (name:
+              let tpl = config.sops.templates.${name};
+              in ''
+                mkdir -p "${dirOf tpl.path}"
+                ${config.sops.substituteCmd} ${tpl.file} > ${tpl.path}
+                chmod "${tpl.mode}" "${tpl.path}"
+                chown "${tpl.owner}" "${tpl.path}"
+                chgrp "${tpl.group}" "${tpl.path}"
```
Mic92, line 104: That might leak the secret for small amount of time when permissions are not applied corretly yet. Better to use the `install` command here to make this step atomically in one command.

Mic92: install -D -m "${tpl.mode}" -o "${tpl.owner}" -g "${tpl.group}" "${tpl.file}" "${tpl.path}"

modules/sops/templates/subs.py
Mic92: mhm. might be type checked anyway but I usually prefer not to introduce global variable as it might become ugly as the program grows.

modules/sops/templates/default.nix
```diff
@@ -0,0 +1,108 @@
+{ config, pkgs, lib, options, ... }:
+with lib;
+with lib.types;
+with builtins;
+let
+  cfg = config.sops;
+  secretsForUsers = lib.filterAttrs (_: v: v.neededForUsers) cfg.secrets;
+  users = config.users.users;
+  substitute = pkgs.writers.writePython3 "substitute" { }
+    (replaceStrings [ "@subst@" ] [ "${subst-pairs}" ] (readFile ./subs.py));
```
Mic92, line 10: This requires an extra derivation per substitute, would it not be better to just pass an argument to `substitute`?

modules/sops/templates/default.nix
Mic92: Mhm, since this is using `stdout` this is not so easy to replace with the install command. Maybe the substitution command should also install the secret? Otherwise we also need to use `(umask 077; <yourcommand> > secret)` to make sure that no other process can read the secret when it's written.

Mic92:
cc @polygon

modules/sops/templates/default.nix
Ninlives: I agree that just pass an argument would be better, I really can't recall why I choose this weird method when writing this piece of code years ago.

modules/sops/templates/default.nix
Ninlives: Is this flow secure enough for creating a secret file:
1. umask 077
2. write the file
3. set group
4. set owner
5. set mode

Ninlives: What if touch the file and set owner & permission before writing to it?

bjornfor: Touch and set perms afterwards is racy. Someone could open the file before the perms are restricted.

Mic92: That's why `umask` would first make the file not readable by any non-root process.

sg-qwt:
It'd also be nice if `sops.templates` could also support **restartsUnits** and **reloadUnits** same as `sops.secrets`. In current sops.templates implantation, I often find myself restarting service because sops.secrets is not changed, but the templates changes often. I can raise a separate issue to log if this PR goes in.

Ninlives:
> It'd also be nice if `sops.templates` could also support **restartsUnits** and **reloadUnits** same as `sops.secrets`. In current sops.templates implementation, I often find myself restarting service because sops.secrets is not changed, but the templates changes often. I can raise a separate issue to log if this PR goes in.

Wouldn't `systemd.services.<name>.reloadTriggers/restartTriggers` help in such cases?

modules/sops/templates/default.nix
```diff
@@ -0,0 +1,103 @@
+{ config, pkgs, lib, options, ... }:
+with lib;
+with lib.types;
+with builtins;
+let
+  cfg = config.sops;
+  secretsForUsers = lib.filterAttrs (_: v: v.neededForUsers) cfg.secrets;
+  users = config.users.users;
```
SuperSandro2000, line 8:
```suggestion
```

modules/sops/templates/default.nix
```diff
@@ -0,0 +1,103 @@
...
+      owner = mkOption {
+        type = str;
+        default = "root";
+        description = ''
+          User of the file.
+        '';
+      };
+      group = mkOption {
+        type = str;
+        default = users.${config.owner}.group;
```
SuperSandro2000, line 56:
```suggestion
        default = config.users.users.${config.owner}.group;
```
we don't need to clutter the scope if it is only used once

modules/sops/templates/default.nix
```diff
@@ -0,0 +1,103 @@
...
+      sops.placeholder = mapAttrs
+        (name: _: mkDefault "<SOPS:${hashString "sha256" name}:PLACEHOLDER>")
+        config.sops.secrets;
+
+      system.activationScripts.renderSecrets = mkIf (cfg.templates != { })
+        (stringAfter ([ "setupSecrets" ]
+          ++ optional (secretsForUsers != { }) "setupSecretsForUsers") ''
+            echo Setting up sops templates...
+            ${concatMapStringsSep "\n" (name:
+              let tpl = config.sops.templates.${name};
```
SuperSandro2000, line 93:
```suggestion
              let
                tpl = config.sops.templates.${name};
                substitute = pkgs.writers.writePython3 "substitute" { } (readFile ./subs.py);
                subst-pairs = pkgs.writeText "pairs" (concatMapStringsSep "\n" (name:
                  "${toString config.sops.placeholder.${name}} ${
                    config.sops.secrets.${name}.path
                  }") (attrNames config.sops.secrets));
```

modules/sops/templates/default.nix
```diff
@@ -0,0 +1,103 @@
...
+with builtins;
+let
+  cfg = config.sops;
+  secretsForUsers = lib.filterAttrs (_: v: v.neededForUsers) cfg.secrets;
+  users = config.users.users;
+  substitute = pkgs.writers.writePython3 "substitute" { } (readFile ./subs.py);
+  subst-pairs = pkgs.writeText "pairs" (concatMapStringsSep "\n" (name:
+    "${toString config.sops.placeholder.${name}} ${
+      config.sops.secrets.${name}.path
+    }") (attrNames config.sops.secrets));
```
SuperSandro2000, line 13:
```suggestion

```

modules/sops/templates/default.nix
```diff
@@ -0,0 +1,103 @@
...
+  coercibleToString = mkOptionType {
+    name = "coercibleToString";
+    description = "value that can be coerced to string";
+    check = strings.isCoercibleToString;
+    merge = mergeEqualOption;
+  };
+  templateType = submodule ({ config, ... }: {
```
SuperSandro2000, line 20: Those two could be inlined, too

SuperSandro2000 (commented):
Why not use pkgs.replace-secret so that we can avoid a runtime dependency on python.

Ninlives:
> Why not use pkgs.replace-secret so that we can avoid a runtime dependency on python.

Oh I have not heard of this package before, but doesn't it also depend on python..?

Mic92:
> Oh I have not heard of this package before, but doesn't it also depend on python..?

It does. I don't have a preference really.

Mic92:
Api/Implementation looks good now. The only thing I would appreciate is extending one of our nixos tests to test the template: https://github.com/Mic92/sops-nix/blob/d8827a8368c307fbc6ed594c9a31d619e7360bed/pkgs/sops-install-secrets/nixos-test.nix#L100 to avoid regressions in the future.

Ninlives:
ping @Mic92

Ninlives:
Any updates here?

modules/sops/templates/default.nix
```diff
@@ -0,0 +1,102 @@
...
+                subst-pairs = pkgs.writeText "pairs" (concatMapStringsSep "\n"
+                  (name:
+                    "${toString config.sops.placeholder.${name}} ${
+                      config.sops.secrets.${name}.path
+                    }") (attrNames config.sops.secrets));
+              in ''
+                mkdir -p "${dirOf tpl.path}"
+                (umask 077; ${substitute} ${tpl.file} ${subst-pairs} > ${tpl.path})
+                chmod "${tpl.mode}" "${tpl.path}"
+                chown "${tpl.owner}:${tpl.group}" "${tpl.path}"
```
Mic92, line 98: I think this bit could be probably just moved into the python script later.
But it'll work as a first attempt.

Mic92:
bors merge

modules/sops/templates/default.nix
```diff
@@ -0,0 +1,102 @@
...
+            '';
+          };
+          owner = mkOption {
+            type = str;
+            default = "root";
+            description = ''
+              User of the file.
+            '';
+          };
+          group = mkOption {
```
mlyxshi, line 45: Can we set default group to `root`?
```
   error: attribute 'users' missing

       at /nix/store/pdi20jccn9ah62y3ny1dqwaqk869x995-source/modules/sops/templates/default.nix:47:23:

           46|             type = str;
           47|             default = config.users.users.${config.owner}.group;
             |                       ^
           48|             description = ''
Build Failed
```

Ninlives: This is in line with the `sops.secrets` settings. Why is `config.users` missing? How do you use `sops-nix`?

Ninlives: Oh, I see the error here, I'll submit a fix soon.

modules/sops/templates/default.nix
Ninlives: The `config` in submodule shadows the toplevel `config`, so this is actually needed.

PR #300, opened 2023-03-17
fix makeSetupHook deprecations

Tomaszal:
From [nixpkgs](https://github.com/NixOS/nixpkgs/blob/c333779d748a996a9a305b3d1fcf6b12b51564ed/pkgs/build-support/trivial-builders.nix#L598):

> calling makeSetupHook without passing a name is deprecated.

> 'deps' argument to makeSetupHook is deprecated and will be removed in release 23.11., Please use propagatedBuildInputs instead.

### DIFF
--- a/pkgs/sops-import-keys-hook/default.nix
+++ b/pkgs/sops-import-keys-hook/default.nix
@@ -1,8 +1,9 @@
 { stdenv, makeSetupHook, gnupg, sops, nix }:
 
 (makeSetupHook {
+  name = "sops-import-keys-hook";
   substitutions = {
     gpg = "${gnupg}/bin/gpg";
   };
-  deps = [ sops gnupg ];
+  propagatedBuildInputs = [ sops gnupg ];
 } ./sops-import-keys-hook.bash)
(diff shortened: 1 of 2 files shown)


Mic92:
bors merge

pkgs/sops-import-keys-hook/default.nix
```diff
@@ -1,8 +1,9 @@
 { stdenv, makeSetupHook, gnupg, sops, nix }:
 
 (makeSetupHook {
+  name = "sops-import-keys-hook";
   substitutions = {
     gpg = "${gnupg}/bin/gpg";
   };
-  deps = [ sops gnupg ];
+  propagatedBuildInputs = [ sops gnupg ];
```
Mic92, line 8: Looks like this breaks nixos 22.11?

Tomaszal: Yeah, not sure what's the best way to go about this deprecation, as `propagatedBuildInputs` does not exist in earlier nixpkgs.

Mic92: @Tomaszal do you want to implement this change?

Mic92:
https://github.com/Mic92/sops-nix/pull/326

Issue #284, opened 2023-02-28
%r not being replaced when using the home-manager module


jghauser:
I've been using the recently introduced home-manager module and it works very nicely except for one snag. 
Say, I try to make vdirsyncer get a password from the sops file, so I add the following:
```nix
{ config, pkgs, lib, ... }:
{
  sops = {
    secrets."vdirsyncer/[redacted]" = { };
  };
  home.packages = [ pkgs.vdirsyncer ];

  xdg.configFile."vdirsyncer/config".text = ''
  ...
  password.fetch = [ "command", "${pkgs.coreutils}/bin/cat", "${config.sops.secrets."vdirsyncer/[redacted]".path}" ]
  ...
  '';
}
```

Additionally, my `home.nix` has the following snippet:
```nix
  sops = {
    age.keyFile = "/path/to/identity/file";
    defaultSopsFile = ../secrets.yaml;
  };
```

Now, if I run vdirsyncer, I get the following error:
```
Feb 26 18:48:47 tiny vdirsyncer[199212]: error: Unknown error occurred: Command '('/nix/store/gn5515zj8skk23jvrrljirgxr10fw9az-coreutils-9.1/bin/cat', '%r/secrets/vdirsyncer/[redacted]')' returned non-zero exit status 1.
```

If I tell vdirsyncer to cat the relevant file directly (`/run/user/1000/secrets/vdirsyncer/[redacted]`), things work just fine. The same problem arises for other software (e.g. borgmatic).

I'm pretty new to nixos, so this might very well be a mistake on my part! Am I missing something?

Thanks a lot for all the work on this very useful piece of software :)

Mic92:
cc @pogobanane

Mic92:
I don't think it's currently possible to know the uid of user in home-manager to compute this path.

jghauser:
The path is stored in my `$XDG_RUNTIME_DIR`:
```
> echo $XDG_RUNTIME_DIR
/run/user/1000
```
Is the problem that the sops-nix service doesn't have access to this?

Mic92:
The problem is that we don't have the actual value of $XDG_RUNTIME_DIR when the home-manager configuration is built. I think some tools might be not so happy either if they receive environment variables in their configuration.

pogobanane:
we choose $XDG_RUNTIME_DIR as the default place to write/link secrets to. If this doesnt work for you, you can set your own absolute path as described in [1]:
[...]

PR #305, opened 2023-03-20
HM: make `secretsMountPoint` and `symlinkPath` configurable

MidAutumnMoon:
Related: https://github.com/Mic92/sops-nix/issues/284

The `%r` specifier used in the home-manager module has been deprecated by systemd for quite a while, and doesn't work smoothly with `AssertPath...` directives in my experiments.

This PR adds two options `defaultSymlinkPath` and `defaultSecretsMountPoint` to make their corresponding manifest field configurable, but their defaults remain intact.

Tested on my laptop setup.

### DIFF
--- a/modules/home-manager/sops.nix
+++ b/modules/home-manager/sops.nix
@@ -25,7 +25,7 @@ let
 
       path = lib.mkOption {
         type = lib.types.str;
-        default = "%r/secrets/${name}";
+        default = "${cfg.defaultSymlinkPath}/${name}";
         description = ''
           Path where secrets are symlinked to.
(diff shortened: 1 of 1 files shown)


Shawn8901:
Hi, as i was roaming around #284, was there a specific reason that you close your PR or did you just lost interest in it (as you also removed your fork).

Shawn8901:
@MidAutumnMoon thank you for the response, do you mind if I resubmit your change? Otherwise the HM module in non-standalone is quite non-functional.

Mic92:
Sorry. Too many projects to cycle through.

PR #318, opened 2023-04-18
Fix scope in `sops.templates`

Ninlives:
The `config` argument in submodule shadows the toplevel `config`, which breaks the default settings.
Also add test for testing default settings.


sg-qwt (approved):
Yes, this fixed template with default group, otherwise failed to build.

Mic92:
bors merge

Mic92:
Would be also nice to mention templates in our README.

PR #328, opened 2023-05-08
Allow nested secrets in secrets.json

srid:
Prior to this change, if `secrets.json` had nested secrets ([example](https://github.com/srid/nixos-config/blob/15d848f475efc505107d3faa78c9ab2125e403cd/secrets.json#L8)) we would see this error:

```
sops-install-secrets: Manifest is not valid: secret jenkins-nix-ci/cachix-auth-token/description in /nix/store/wxm763za3rbrpiijfbgss9g5ll0sd29z-secrets.json is not valid: Key 'jenkins-nix-ci' does not refer to a dictionary
```

The reason happens to be that introspecting the map key to be `interface` fails, when it is in fact a string. This PR makes it so that we always expect the key to be a string (what else could it be?). It also improves the error message, by telling the user what the actual value type is.

### DIFF
--- a/pkgs/sops-install-secrets/main.go
+++ b/pkgs/sops-install-secrets/main.go
@@ -10,6 +10,7 @@ import (
 	"os/user"
 	"path"
 	"path/filepath"
+	"reflect"
 	"strconv"
 	"strings"
 	"syscall"
@@ -239,13 +240,13 @@ func recurseSecretKey(keys map[string]interface{}, wantedKey string) (string, er
 		if !ok {
 			return "", fmt.Errorf("The key '%s' cannot be found", keyUntilNow)
 		}
-		valWithWrongType, ok := val.(map[interface{}]interface{})
+		valWithWrongType, ok := val.(map[string]interface{})
 		if !ok {
-			return "", fmt.Errorf("Key '%s' does not refer to a dictionary", keyUntilNow)
+			return "", fmt.Errorf("Expected key '%s' to refer to a dictionary, but it is a '%s' instead", keyUntilNow, reflect.TypeOf(val).Elem())
 		}
 		currentData = make(map[string]interface{})
 		for key, value := range valWithWrongType {
-			currentData[key.(string)] = value
+			currentData[key] = value
 		}
 	}
 


srid:
Huh, this is causing the yaml processing to regress.

pkgs/sops-install-secrets/main.go
```diff
@@ -239,13 +240,25 @@ func recurseSecretKey(keys map[string]interface{}, wantedKey string) (string, er
...
-		valWithWrongType, ok := val.(map[interface{}]interface{})
-		if !ok {
-			return "", fmt.Errorf("Key '%s' does not refer to a dictionary", keyUntilNow)
-		}
 		currentData = make(map[string]interface{})
-		for key, value := range valWithWrongType {
-			currentData[key.(string)] = value
+		// The 'if' here is to deal with key type discrepancy between YAML and
+		// JSON.  With YAML, it is 'interface {}'; with JSON, it is 'string'.
+		if format == Json {
```
Mic92, line 246: Looks good. Can you also extend one of our tests to have a nested key?

srid: @Mic92 What's the best way to do that? I don't really understand the test infrastructure in this repo. Especially what's going on with the Go tests. `nixos-tests.nix` seems more pratical; but how do you edit `secrets.json`? Running `nix run . pkgs/sops-install-secrets/test-assets/secrets.json` throws:

```
fingerprint: 26F82B82FDFFA024E08B9C8B67936C83AAC837D4
mv: cannot stat '/root/.gnupg': Permission denied
```

shivaraj-bh:
I am trying to write the test for this. Here's how the `nixos-test.nix` looks:
```nix
nested-json = makeTest {
    name = "sops-nested-json-secrets";
    nodes.server = {
      imports = [ ../../modules/sops ];
      sops = {
        age.keyFile = ./test-assets/age-keys.txt;
        defaultSopsFile = ./test-assets/secrets.json;
        secrets."nested/test/file" = { };
      };
    };

    testScript = ''
      start_all()
      server.succeed("cat /run/secrets/nested/test/file | grep -q 'another value'")
    '';
  } {
    inherit pkgs;
    inherit (pkgs) system;
  };
```
`Expected`: `testScript` should fail with `Key 'nested' does not refer to a dictionary`
`Actual`: Test passes
On the other hand, if I try to run `nix build .#nixosConfigurations.actual.config.system.build.toplevel` on @srid 's [nixos-config](https://github.com/srid/nixos-config) with `sops-nix.url` pointing to `github:Mic92/sops-nix`, I am able to reproduce the same error. 
@Mic92 Is there something that's happening under-the-hood in `nixos-test.nix` that I might be missing?

OsiPog:
How is the state on this PR? I would like to use nested JSON for my secrets. Can I help out here?

Issue #120, opened 2021-09-29
Support mounting yaml and json files as is?


ljani:
My current use case: I'd like to pass a `sops` encrypted `.yaml` configuration file to my app after decryption as is instead of picking a single key from the file. So, `sops-nix` is a little too clever for my use case and it should focus only on decrypting the file, not parsing it.

In other words, I have this file `./secrets/my-config.yaml` in my flake:

```yaml
my-secret1: ENC[AES256_GCM,data:tkyQPQODC3g=,iv:yHliT2FJ74EtnLIeeQtGbOoqVZnF0q5HiXYMJxYx6HE=,tag:EW5LV4kG4lcENaN2HIFiow==,type:str]
my-secret2: ENC[AES256_GCM,data:tkyQPQODC3g=,iv:yHliT2FJ74EtnLIeeQtGbOoqVZnF0q5HiXYMJxYx6HE=,tag:EW5LV4kG4lcENaN2HIFiow==,type:str]
sops:
    kms: []
    gcp_kms: []
    azure_kv: []
    hc_vault: []
...
```

It should be mounted as `/run/secrets/my-config.yaml`:

```yaml
my-secret1: hello
my-secret2: hello
```

Then I could pass the `/run/secrets/my-config.yaml` to my app to read any secrets from.

I could do this using a `format=binary` file, but then I'd lose eg. syntax highlight based on the file extension and have to pass use `format=binary` everywhere. Alternatively, I could use a literal multi line yaml string, but then again I'd lose highlight.

I wonder if this is already supported, because the `README.md` [mentions `home-assistant-secrets.yaml`](https://github.com/Mic92/sops-nix/blob/ae84c313c5250a832d61dae9e1e659b27542c47b/README.md#L558), which is similar to what I'm looking for? But I think that was defined as something like this:

```yaml
home-assistant-secrets.yaml: |
    my-secret1: hello
    my-secret2: hello
```

(Notice how the renderer fails to highlight the yaml keys `my-secret1` and `my-secret2`).

Does this make any sense to you?

Mic92:
Currently I don't see any other way but using either the multi-line variant or encrypting the yaml file as a binary, in which case you would still not be able to use sops as an editor of the file.
Personally I use multline yaml values to encode yaml secrets.

ljani:
> Notice that this a limitation in sops itself, where sops-nix has little control over.

[...]

PR #338, opened 2023-05-30
Emit plain file when key is empty

slaier:
Like `sops --extract ''`, get the plain file instead of an single key.
Fixes: #120

### DIFF
--- a/modules/home-manager/sops.nix
+++ b/modules/home-manager/sops.nix
@@ -15,7 +15,7 @@ let
 
       key = lib.mkOption {
         type = lib.types.str;
-        default = name;
+        default = if cfg.defaultSopsKey != null then cfg.defaultSopsKey else name;
         description = ''
           Key used to lookup in the sops file.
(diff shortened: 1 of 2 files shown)


thomaslepoix:
I link this here so that it won't be missed : https://github.com/slaier/sops-nix/pull/1

metaspace:
Any reason this was dropped?

thomaslepoix:
I think @slaier simply removed its own fork without having this PR in mind, as some months have passed in between.

But it is still relevant and should be reopen.

Just as a stop gap, it can be applied like this in a sops-nix `overrideAttrs` :

```nix
patches = [
  (fetchpatch {
    url = "https://github.com/Mic92/sops-nix/pull/338.patch";
    hash = lib.fakeHash;
  })
];

```

thomaslepoix:
Oops sorry, it seems code has diverged too much to apply the patch as is. I can take a look to revive it.

PR #346, opened 2023-06-16
Configure the systemd user service to start with graphical session if use of a passphrase is detected

physics-enthusiast:
If GnuPG is enabled as a key source, it must start only after the display manager starts, else the pinentry prompt where the user is asked for the passphrase errors out with `error calling pinentry: No such file or directory <Pinentry>`. Hence, if the config option `gnupg.home` is set, `sops-nix.service` should startup on `graphical-session.target` rather than `default.target`.


Mic92:
Thanks!

PR #352, opened 2023-06-21
fix: darwin setup on ventura

roman:
### Context

I discovered a few issues when running nix-sops on macOS Ventura; each commit has some documentation explaining its rationale.


Mic92:
@mergify queue

Mic92:
Thanks

Issue #236, opened 2022-10-09
sops-install-secrets: support specifying custom fstype


caguiclajmg:
Deploying sops-nix to an OpenVZ host running kernel 4.19.0 fails due to lack of support for `ramfs`, would be nice to allow [specifying the fs type](https://github.com/Mic92/sops-nix/blob/912f9ff41fd9353dec1f783170793699789fe9aa/pkgs/sops-install-secrets/main.go#L322) (e.g. `tmpfs` or even plain dir) as a flag in `sops-install-secrets` as well as an option in the `sops-nix` module.

I do understand the security implications even with `tmpfs` as you risk getting the contents swapped to disk.

I'm writing the patch myself as I have no way around this one but still want to keep using sops-nix for secret management, would just like to know if there is interest in upstream accepting the patch.

Mic92:
I would accept a patch that add an option with a proper description about the security implication.

KiaraGrouwstra:
should be fixed by #355

PR #355, opened 2023-07-03
Add configuration option to use tmpfs in place of ramfs

wrmilling:
Taking a stab at an implementation to address #236. 

This PR adds an option to use tmpfs in place of ramfs on systems where ramfs is unavailable. Happy to workshop the warning if the verbiage is not clear enough. 

Tested on two systems: personal laptop used ramfs as expected (new option left default), OpenVZ server used tmpfs as configured (`sops.useTmpfs = true;`).


modules/sops/default.nix
```diff
@@ -242,6 +243,17 @@ in {
...
     };
 
+    useTmpfs = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Use tmpfs in place of ramfs for secrets storage.
+
+        *WARNING*
+        Enabling this option has the potential to write secrets to disk unencrypted if the tmpfs volume is written to swap. Do not use unless absolutely necessary.
```
Mic92, line 253:
```suggestion
        Enabling this option has the potential to write secrets to disk unencrypted if the tmpfs volume is written to swap. Do not use unless absolutely necessary.
        
        Consider enabling swap encryption when using this option:
        
        ```
        swapDevices = [{
          device = "/dev/sdXY";
          randomEncryption.enable = true; 
        }];
        ```
```

modules/sops/default.nix
```diff
@@ -242,6 +243,17 @@ in {
       '';
     };
 
+    useTmpfs = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
```
Mic92, line 249:
```suggestion
      description = lib.mkDoc ''
```

Mic92:
@mergify squash

Mic92:
@mergify queue

PR #360, opened 2023-07-10
Add some helpful tips to the README

anoadragon453:
A couple notes that tripped me up when initially trying this. I hope they can help make future journeys smoother :)

- Getting the syntax of `.sops.yaml` wrong can cause vague errors when it comes to encrypting secrets files.
- `path_regex` needs to include all file extensions that you intend to encrypt. `sops` supports YAML, JSON, ENV, INI and binary files, so suggesting that those be picked up by default seems sensible. Personally, I had trouble figuring out why `sops` wouldn't accept my `.env` file - again with a confusing error message.

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,10 +232,11 @@ creation_rules:
       - *admin_bob
 ```
 
-**Note:** be sure to not include a `-` before subsequent key types under `key_groups`
-(i.e. `age` in the above example should not have a `-` in front). If you do, `sops`
-will produce [a vague error](https://github.com/getsops/sops/issues/857) when
-encrypting files.
+**Note:**
+Be sure to not include a `-` before subsequent key types under `key_groups`
+(i.e. `age` in the above example should not have a `-` in front).
+This will otherwise cause sops to require multiple keys (shamir secret sharing)
+to decrypt a secret, which breaks normal sops-nix usage.
 
 </details>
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -224,14 +224,19 @@ keys:
...
       - *admin_alice
       age:
       - *admin_bob
 ```
 
+**Note:** be sure to not include a `-` before subsequent key types under `key_groups`
+(i.e. `age` in the above example should not have a `-` in front). If you do, `sops`
+will produce [a vague error](https://github.com/getsops/sops/issues/857) when
+encrypting files.
+
```
Mic92, line 239: I think this is a common pitfall even I went into a couple of times.
I would extends this a bit to capture more what is actually the difference between the two is. What do you think about the following?

> **Note** be sure to not include a `-` before subsequent key types under `key_groups` (i.e. `age` in the above example should not have a `-` in front).
This will otherwise cause sops to require multiple keys (shamir secret sharing) to decrypt a secret, which breaks normal sops-nix usage.

anoadragon453: Oh TIL that syntax becomes rendered in GitHub-flavoured markdown! And that it was not an invalid configuration I did, but just one that I didn't intend.

I like your suggestion, and have applied it in e4e1a9f.

Mic92:
Thanks for the documentation contributions!
@mergify queue

PR #384, opened 2023-08-15
fix: add missing argument for MountSecretFs on darwin

### LINKED ISSUE #383: Does not compile anymore on Mac
Since https://github.com/Mic92/sops-nix/commit/339a5594021a3c1bd1fb3adfd967820eceb71a95 does it not compile anymore on MacOS.

```
       > Building subPackage ./pkgs/sops-install-secrets
       > # github.com/Mic92/sops-nix/pkgs/sops-install-secrets
       > pkgs/sops-install-secrets/main.go:937:82: too many arguments in call to MountSecretFs
       >      have (string, int, bool, bool)
       >         want (string, int, bool)
```

The fourth argument is missing in the darwin function 😅

shyim:
fixes #383

### DIFF
--- a/pkgs/sops-install-secrets/darwin.go
+++ b/pkgs/sops-install-secrets/darwin.go
@@ -53,7 +53,7 @@ func SecureSymlinkChown(symlinkToCheck string, expectedTarget string, owner, gro
 // mydev=`hdiutil attach -nomount ram://$NUMSECTORS`
 // newfs_hfs $mydev
 // mount -t hfs $mydev /tmp/mymount
-func MountSecretFs(mountpoint string, keysGid int, useTmpfs bool, userMode bool) error {
+func MountSecretFs(mountpoint string, keysGid int, _useTmpfs bool, userMode bool) error {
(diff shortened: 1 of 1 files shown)


peanutbother:
this silences the error as now the function signature is the right one, but why was the call changed in the first place and why is the parameter not used?

pkgs/sops-install-secrets/darwin.go
```diff
@@ -53,7 +53,7 @@ func SecureSymlinkChown(symlinkToCheck string, expectedTarget string, owner, gro
 // mydev=`hdiutil attach -nomount ram://$NUMSECTORS`
 // newfs_hfs $mydev
 // mount -t hfs $mydev /tmp/mymount
-func MountSecretFs(mountpoint string, keysGid int, userMode bool) error {
+func MountSecretFs(mountpoint string, keysGid int, useTmpfs bool, userMode bool) error {
```
allthatjazzleo, line 56:
```suggestion
func MountSecretFs(mountpoint string, keysGid int, _useTmpfs bool, userMode bool) error {
```
maybe add underscore to indicate it is not used

PR #388, opened 2023-08-23
fix templates group default

mrVanDalo:


mrVanDalo:
Not tests, and just seeing that you defined `users` in the let block. But I'm getting this error:
```

```

PR #389, opened 2023-08-24
fix: systemd unit file

Mikilio:
Make sure passwords are ready for services part of graphical-session.target


Mic92:
@mergify rebase

PR #417, opened 2023-10-08
Allow multiple sopsFiles for easier common secrets sharing across multiple configuration

vdbe:
Allow multiple sopsFiles for easier secret sharing across multiple configurations.

For example you have files

secrets/common.yaml
```yaml
hashed_password: ....
vpn_user: ...
vpn_password: ...
vpn_cert: | ...
```
secrets/\<system\>/default.yaml
```yaml
vpn_user: ...
vpn_password: ...
```
with `sops.defaultSopsFiles = [ ./secrets/common.yaml ./secrets/<system>/default.yaml ];` in the configuration for `<system>`.

This would allow overriding of shared secrets across hosts without manually checking and changing the required sopsFile per secret.


vdbe:
Current changes are just a rough draft

Mic92:
Only question is how you order priority between those in case keys overlap? Otherwise it would make sense to throw an error so that users don't accidentally shadow secrets.

vdbe:
My idea was to have the files in increasing priority,
this way you can have a default common file and append system/user specific files.

In other words search the list of sops files in reverse order and use the first found secret.

In the example above vpn_user and vpn_password in secrets/common.yaml would be shadowed by vpn_user and vpn_password in secrets/\<system\>/default.yaml.

Mic92:
It's just hard to enforce an order in lists if you those lists get defined in different nixos modules.

Mic92:
Ah. I think one can use lib.mkBefore for host files.

modules/sops/default.nix
```diff
@@ -77,13 +80,27 @@ let
...
         '';
       };
+      sopsFiles = mkOption {
+        type = types.listOf types.path;
+        defaultText = "\${config.sops.defaultSopsFile}";
+        description = ''
+          Sops file the secret is loaded from.
+        '';
+      };
       sopsFileHash = mkOption {
```
Mic92, line 90: We should deprecated this one than.

modules/sops/default.nix
```diff
@@ -172,6 +189,13 @@ in {
         Default sops file used for all secrets.
       '';
     };
```
Mic92, line 191: And this one with a warning.

modules/sops/default.nix
vdbe: Deprecate sopsFiles? Just added that this PR, do you mean deprecate sopsFile?

Mic92: We shouldn't really need both sopsFileHash and sopsFilesHash, no?

Mic92: They could be point to the same value.

pkgs/sops-install-secrets/main.go
```diff
@@ -494,17 +500,31 @@ func (app *appContext) validateSecret(secret *secret) error {
...
+	for i := len(files) - 1; i >= 0; i-- {
+		err := app.validateSopsFile(secret, &files[i])
+		if err == nil {
+			// Found valid sopsFile
+			break
+		} else if i == 0 {
+			// No valid sopsFile found in sopsFiles
+			return err
+		}
+	}
```
vdbe, line 526: The error "secret %s defined the format of %s as %s, but it was specified as %s in %s before" from `validateSopsFile` is ignored

pkgs/sops-install-secrets/main.go
```diff
@@ -258,40 +258,47 @@ func recurseSecretKey(keys map[string]interface{}, wantedKey string) (string, er
...
+				}
+			case Json:
+				if err := json.Unmarshal(plain, &sourceFile.data); err != nil {
+					return fmt.Errorf("Cannot parse json of '%s': %w", s.SopsFiles[i], err)
+				}
+			default:
+				return fmt.Errorf("Secret of type %s in %s is not supported", s.Format, s.SopsFiles[i])
+			}
+		}
 		switch s.Format {
```
vdbe, line 284: Assuming all sopsFiles have the same format

vdbe (commented):
Something like this?

pkgs/sops-install-secrets/main.go
```diff
@@ -393,40 +400,40 @@ func lookupKeysGroup() (int, error) {
   return 0, fmt.Errorf("Can't find group 'keys' nor 'nogroup' (%w).", err2)
 }
 
-func (app *appContext) loadSopsFile(s *secret) (*secretFile, error) {
+func (app *appContext) loadSopsFile(s *secret, sopsFileIndex int) (*secretFile, error) {
```
Mic92, line 403: Can we not pass in the sopsFile as a pointer here rather than using an index?

pkgs/sops-install-secrets/main.go
```diff
@@ -494,17 +500,31 @@ func (app *appContext) validateSecret(secret *secret) error {
 		return fmt.Errorf("Unsupported format %s for secret %s", secret.Format, secret.Name)
 	}
 
-	file, ok := app.secretFiles[secret.SopsFile]
-	if !ok {
-		maybeFile, err := app.loadSopsFile(secret)
-		if err != nil {
-			return err
+	files := []secretFile {};
+	for index, sopsFile := range secret.SopsFiles {
```
Mic92, line 504: You already have you reference here.

pkgs/sops-install-secrets/main.go
Mic92: Mhm. I would build a new error message that would give you all files where the secret has been looked for instead.

modules/sops/default.nix
```diff
@@ -1,18 +1,21 @@
...
-  secretType = types.submodule ({ config, ... }: {
+  secrets = mapAttrs (_: secret: removeAttrs secret ["sopsFile"]) cfg.secrets;
+  regularSecrets = lib.filterAttrs (_: v: !v.neededForUsers) secrets;
+  secretsForUsers = lib.filterAttrs (_: v: v.neededForUsers) secrets;
+  secretType = types.submodule ({ config, options, ... }: {
     config = {
-      sopsFile = lib.mkOptionDefault cfg.defaultSopsFile;
-      sopsFileHash = mkOptionDefault (optionalString cfg.validateSopsFiles "${builtins.hashFile "sha256" config.sopsFile}");
+      sopsFile = mkOptionDefault cfg.defaultSopsFile;
+      sopsFiles = if options.sopsFile.isDefined then warn "`sops.secrets.<name>.sopsFile` is being deprecated, use `sops.secrets.<name>.sopsFiles` instead" [ config.sopsFile ] else (lib.mkOptionDefault cfg.defaultSopsFiles);
```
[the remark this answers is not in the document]
Mic92: Can we not add `defaultSopsFile` to the new `defaultSopsFiles` list?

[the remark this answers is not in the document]
vdbe: Would it not be better to keep `defaultSopsFile` and `defaultSopsFiles` separate and place `sopsFile` before `sopsFiles` per secret?
If you merge in default you would need to remove `sopsFile` 
otherwise `sopsFiles` would become [ `sopsFile` `defaultSopsFile` ] ++ `defaultSopsFiles`.

The only way I was able to implement this feels however a bit hacky

```nix
    config = mkMerge [{
      sopsFile = mkOptionDefault cfg.defaultSopsFile;
      sopsFiles = mkOptionDefault cfg.defaultSopsFiles;
      sopsFilesHash = mkOptionDefault (optionals cfg.validateSopsFiles (forEach config.sopsFiles (builtins.hashFile "sha256")));
    } 
    {
      sopsFiles = mkIf (config.sopsFile != null) (mkOverride options.sopsFile.highestPrio (mkBefore [config.sopsFile]));
    }];
```

The effect of this is that when both `sopsFile` and `sopsFiles` are defined (and not null) with different priorities only one will make it into `config.sopsFiles`

gabevenberg:
is this dead in the water? is there another way to share secrets among hosts without copying secrets between individual hosts secrets.yaml files?

fabiosouzadev (approved):

PR #421, opened 2023-10-14
don't substitute binaries

edwtjo:
currently mixing templates and binary keys in the same configuration b0rks evaluation since `f.read()` speaks utf-8, it also doesn't make sense to template binaries which are not string encoded


Mic92:
Independent of that we should probably don't treat secrets as string but as opaque bytes, but let's leave this to a future PR.

Mic92:
@mergify queue

PR #433, opened 2023-11-03
Rename passwordFile to hashedPasswordFile

Pizmovc:
This follows the rename that was done in nixpkgs.

Reference PR https://github.com/NixOS/nixpkgs/pull/254080

### DIFF
--- a/pkgs/sops-install-secrets/nixos-test.nix
+++ b/pkgs/sops-install-secrets/nixos-test.nix
@@ -36,7 +36,7 @@
 
       users.users.example-user = {
         isNormalUser = true;
-        passwordFile = config.sops.secrets.test_key.path;
+        hashedPasswordFile = config.sops.secrets.test_key.path;
       };
     };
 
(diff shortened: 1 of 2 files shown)


pkgs/sops-install-secrets/nixos-test.nix
```diff
@@ -36,7 +36,7 @@
 
       users.users.example-user = {
         isNormalUser = true;
-        passwordFile = config.sops.secrets.test_key.path;
+        hashedPasswordFile = config.sops.secrets.test_key.path;
```
Mic92, line 39: For the 22.03 test we need to make this conditional.

Pizmovc: Ah yes, of course. I've added a fixup [b43feda](https://github.com/Mic92/sops-nix/pull/433/commits/b43feda77e83656972d2b5e8d4c0d36f74cd8618).

Mic92:
@mergify queue

Pizmovc:
@Mic92 , not sure if you noticed or even if you care about this, but commits did not get squashed before being merged, so now there are two commits on the master branch when there should only be one fixup-ed commit.

Mic92:
Missed it. I can probably use mergify next time to squash it.

PR #437, opened 2023-11-13
sops-install-secrets: don't trigger a rebuild when flake.lock changes

SuperSandro2000:
When you have a follow in your flake inputs, sops-install-secrets is rebuild everytime the flake.lock changes despite that being a noop. When filtering src this can be avoided.


Mic92:
@mergify queue

PR #454, opened 2023-12-13
marshal nested secrets value to string

iosmanthus:
This pull request marshals nested secret values according to the format of the secret file.

### DIFF
--- a/pkgs/sops-install-secrets/main.go
+++ b/pkgs/sops-install-secrets/main.go
@@ -209,7 +209,7 @@ type plainData struct {
 	binary []byte
 }
 
-func recurseSecretKey(keys map[string]interface{}, wantedKey string) (string, error) {
+func recurseSecretKey(format FormatType, keys map[string]interface{}, wantedKey string) (string, error) {
 	var val interface{}
 	var ok bool
 	currentKey := wantedKey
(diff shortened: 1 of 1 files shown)


iosmanthus:
Seem the serialized yaml value has some newline, I should trim the output string.

Mic92:
So the idea is that we can have yaml and json encrypted without having to put them in strings? Would be nice to extend the test case those those as well: https://github.com/Mic92/sops-nix/blob/master/pkgs/sops-install-secrets/nixos-test.nix

The script to update secrets is here: https://github.com/Mic92/sops-nix/blob/master/pkgs/sops-install-secrets/test-assets/sops-edit

iosmanthus:
> So the idea is that we can have yaml and json encrypted without having to put them in strings? Would be nice to extend the test case those those as well: [`master`/pkgs/sops-install-secrets/nixos-test.nix](https://github.com/Mic92/sops-nix/blob/master/pkgs/sops-install-secrets/nixos-test.nix?rgh-link-date=2023-12-13T17%3A01%3A13Z)
> 
> The script to update secrets is here: [`master`/pkgs/sops-install-secrets/test-assets/sops-edit](https://github.com/Mic92/sops-nix/blob/master/pkgs/sops-install-secrets/test-assets/sops-edit?rgh-link-date=2023-12-13T17%3A01%3A13Z)

The basic idea is to support any value in a nested value.  However, I found some edge cases for this pull request:
1. It supports secrets value to be a number but shows as a string in the template rendering, I might add some type of information into the placeholder.
2. If the secret is string formed by numbers, like "113", the rendered template would be like: ""123"".


I would add some tests to `nixos-test.nix`, and some unit tests for `recurseSecretKey`.

Mic92:
Maybe it should only render arrays and hashmaps?

iosmanthus:
> Maybe it should only render arrays and hashmaps?

Perhaps we should marshal nested secrets only in JSON format because it is indent-insensitive. The YAML nested secrets are difficult to fit in the template due to indentation issues.

geoffreygarrett:
+1

geoffreygarrett:
Curious how this PR will affect the secrets declaration, I was attempting this before finding this PR:

```nix
sops.secrets."cachix-auth-token" = {
    mode = "0400"; 
    key = "'[\"cachix-auth-token\"][\"value\"]'";
};
```

Which mirrors the CLI `extract`. I can't tell by this PR diff what this will allow, might be useful to add a test & doc update?

PR #455, opened 2023-12-16
Remove confusing and redundant left over text

EmergentMind:
This line is left over from a set of instructions that were previously incorporated into an early console example under "you can generate yourself a key:" above.


Mic92:
@mergify queue

PR #461, opened 2024-01-01
Make symlink path configurable

jdonszelmann:
There used to be a comment here

asking whether this needed to be configurable? Well, I needed it to be configurable. So I made it so. I thought I'd contribute the change upstream. What do you think?

I'm sorry, I alro ran autoformat on the file. Maybe that's nice to do on the entire project, though I can undo it if you really want to.

### DIFF
--- a/modules/sops/default.nix
+++ b/modules/sops/default.nix
@@ -19,7 +19,7 @@ let
         type = types.str;
         default = config._module.args.name;
         description = ''
-          Name of the file used in /run/secrets
+          Name of the file used in /run/secrets (or your configured symlinkPath)
         '';
       };
       key = mkOption {
@@ -153,7 +153,7 @@ let
   # Skip ssh keys deployed with sops to avoid a catch 22
   defaultImportKeys = algo:
     if config.services.openssh.enable then
-      map (e: e.path) (lib.filter (e: e.type == algo && !(lib.hasPrefix "/run/secrets" e.path)) config.services.openssh.hostKeys)
+      map (e: e.path) (lib.filter (e: e.type == algo && !(lib.hasPrefix cfg.symlinkPath e.path)) config.services.openssh.hostKeys)
     else
       [ ];
 in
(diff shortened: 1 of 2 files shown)


modules/sops/default.nix
```diff
@@ -19,7 +19,7 @@ let
         type = types.str;
         default = config._module.args.name;
         description = ''
-          Name of the file used in /run/secrets
+          Name of the file used in /run/secrets (or your configured symlinkPath)
```
SuperSandro2000, line 22:
```suggestion
          Name of the file used in /run/secrets or your configured symlinkPath
```

modules/sops/default.nix
```diff
@@ -33,15 +33,16 @@ let
       };
       path = mkOption {
         type = types.str;
-        default = if config.neededForUsers then "/run/secrets-for-users/${config.name}" else "/run/secrets/${config.name}";
+        default = if config.neededForUsers then "/run/secrets-for-users/${config.name}" else "${cfg.symlinkPath}/${config.name}";
         defaultText = "/run/secrets-for-users/$name when neededForUsers is set, /run/secrets/$name when otherwise.";
         description = ''
           Path where secrets are symlinked to.
           If the default is kept no symlink is created.
+          If cfg.symlinkPath is set, /run/secrets will become that path
```
SuperSandro2000, line 41:
```suggestion
          If `sops.symlinkPath` is set, `/run/secrets` is that path
```
I am not sure if I understand this description correct

modules/sops/default.nix
```diff
@@ -181,6 +182,22 @@ in {
       '';
     };
 
+    secretsMountPoint = mkOption {
+      type = types.str;
+      default = "/run/secrets.d";
+      description = ''
+        The location where the secrets are stored. This is then symlinked to cfg.SymlinkPath (default: /run/secrets)
```
SuperSandro2000, line 189:
```suggestion
        The location where the secrets are stored. This is symlinked to sops.SymlinkPath which defaults to `/run/secrets`.
```

PR #466, opened 2024-01-10
allow ssh key import to fail

Mic92:
We import ssh keys by default if openssh is enabled. However if users are using age keys while using sops to deploy ssh keys we have a catch-22.
While users could use lib.mkForce to empty the list, this is not intuitive


Mic92:
@mergify queue

PR #469, opened 2024-01-15
made sops-install-secrets working when crossSystem is set

dguibert:
In NixOS setting `CrossSystem` allow to compile for a system that can be different of the host system but the tools that need to be run on the build host must come from `pkgs.buildPackages`.


modules/home-manager/sops.nix
```diff
@@ -2,7 +2,7 @@
 
 let
   cfg = config.sops;
-  sops-install-secrets = (pkgs.callPackage ../.. {}).sops-install-secrets;
+  sops-install-secrets = (pkgs.buildPackages.callPackage ../.. {}).sops-install-secrets;
```
Mic92, line 5: Are you sure this correct?
sops-install-secrets needs to run on the host system i.e if you do a x86_64 -> aarch64 build than this would yield an x86_64 binary that cannot run on aarch64.

Mic92: When do you encounter a case where this is needed at build time?

dguibert: Ah right, I'm not sure this is correct.

I encounter this not a build time but at activation time as I deploy a x86_64 node an aarch64 home-manager profile that is cross compiled. In this case, the activation scripts need to be x86_64 instead of aarch64.

See https://github.com/nix-community/home-manager/pull/4907, where I've made a similar PR for home-manager.

This will break more "standard" workflow where the activation is done in the node itself.

May I propose an update with an option like `activationPackageSet` defaulting to `pkgs`?

Mic92: You are deploying an aarch64 profile on a x86_64 node? How do you use aarch64 binaries on x86_64? I don't understand the use case. As the patch is just now it would break cross setups i.e. I have a riscv64 machine that gets build by a x86 machine.

dguibert: My $HOME is on a NFS mount point that is shared between several nodes (x86_64 and aarch64).
As the aarch64 nodes are not much powerfull (or used by other users), my aarch64 profile is build on a x86_64 node and also deployed from it.

NB: On that cluster, binfmt_misc is not configured to run aarch64 binaries on x86_64.

Mic92: Ok. But don't you run the actual activation phase on the aarch64 machine natively?

dguibert: In fact not and that why I made this patch. I'm building and deploying from the build node.

I've updated this patch and its related patch on home-manager at https://github.com/nix-community/home-manager/pull/4907

Mic92: Ok. I can merge this, when home-manager patch is in.

PR #471, opened 2024-01-20
Add info about hash passwords

haru02w:
Since raw passwords don't work, add information about how to create the hashed passwords


Mic92:
@mergify queue

PR #482, opened 2024-02-08
move secrets-fo-users to it's own module

Mic92:
This preparation to support sysusers.
No behavior change.


Mic92:
@mergify queue

PR #484, opened 2024-02-08
make sops-install-secrets work with sysusers

Mic92:


pkgs/sops-install-secrets/nixos-test.nix
```diff
@@ -55,6 +36,38 @@
...
+    '';
+  } {
+    inherit pkgs;
+    inherit (pkgs) system;
+  };
+
+  user-passwords = userPasswordTest "sops-user-passwords" {};
+  user-passwords-sysusers = userPasswordTest "sops-user-passwords-sysusers" {
+    systemd.sysusers.enable = true;
+    users.mutableUsers = true;
```
Mic92, line 66: FIXME: doesn't work with mutableUsers...

modules/sops/default.nix
```diff
@@ -312,8 +314,22 @@ in {
 
       sops.environment.SOPS_GPG_EXEC = lib.mkIf (cfg.gnupg.home != null) (lib.mkDefault "${pkgs.gnupg}/bin/gpg");
 
+      # When using sysusers we no longer no longer be started as an activation script because those are started in initrd while sysusers is started later.
```
SuperSandro2000, line 317: no longer is written twice

dasJ (approved):

Issue #478, opened 2024-02-04
Sops nix as a home-manager module does not create a symlink


ircurry:
I have followed the README instructions, importing sops-nix in home manager via 
```nix
home-manager.sharedModules = [inputs.sops-nix.homeManagerModules.sops]
```
 (as I am using home-manager as a nixos module), set the sops default file, format and key via
```nix
sops = {
  defaultSopsFile = ../../secrets/secrets.yaml;
  defaultSopsFormat = "yaml";
  age.keyFile = "/home/myuser/.config/sops/age/keys.txt";
};
```
and attempted to decrypt it via 
```nix
config = {
  sops.secrets.test = {
    path = "%r/test.txt"; # as well as through other paths such as "/home/recur/test.txt", "$XDG_RUNTIME_DIR/secrets/mySecret", and not setting it at all
  };
};
```

The expected behavior would be that the decrypted file would be put in the specified location. However, no symlink is made, though running 
```bash
sudo cat /run/secrets/test
```
prints the contents of the secret. 

I am unsure if this is a duplicate of #287 or not.

w4tsn:
As I understand it the `home-manager` module creates the secrets under `/run/user/<uid>/secrets`, so the file should be `/run/user/<uid>/secrets/test.txt` after `home-manager switch` - but only after you've executed `systemctl --user (re)start sops-nix.service`

yqlbu:
> As I understand it the `home-manager` module creates the secrets under `/run/user/<uid>/secrets`, so the file should be `/run/user/<uid>/secrets/test.txt` after `home-manager switch` - but only after you've executed `systemctl --user (re)start sops-nix.service`

Hi @w4tsn, thanks for your kind tips. It is indeed the case. After restarting the `sops-nix.service`, I can see the secrets there. Any idea how to automate the restart service step?

floating-cat:
I wasted some time investigating why sops-nix doesn't generate the secret files for me after using `home-manager switch`, then I noticed I needed to start `systemctl start --user sops-nix` once to generate them and found this issue.
I think sops-nix should do it automatically.

Mic92:
I am currently not using the home-manager sops module but I am open to merge fixes.

yqlbu:
> I wasted some time investigating why sops-nix doesn't generate the secret files for me after using `home-manager switch`, then I noticed I needed to start `systemctl start --user sops-nix` once to generate them and found this issue. I think sops-nix should do it automatically.

[...]

PR #485, opened 2024-02-09
Document restarting systemd service when using home-manager module

ircurry:
Document restarting sops-nix.service after updating secrets. This wouldn't solve #478 but does allow the user to work around the issue.


Mic92:
What do you think about the following snippet? https://github.com/Mic92/sops-nix/pull/491/files
Is this good enough?

ircurry:
Works like a charm, thank you for this amazing tool.

PR #491, opened 2024-02-15
add nix config snippet to restart sops-nix service

w4tsn:
As home-manager does not restart the `sops-nix` unit automatically a snippet to instruct home-manager to do so is added.

Home-manager could be instructed to restart the user service from the sops-nix home-manager module instead. Usually home-manager restarts units which changed. Since the sops-nix unit does not change when secrets change this does not trigger automatically.

There are two options:
- let sops-nix home-manager module compute a chained hash over all secrets and place it inside the unit file, so it changes every time the secrets change
- use X-SwitchMethod and X-Restart-Triggers See nix-community/home-manager#3865

### DIFF
--- a/README.md
+++ b/README.md
@@ -800,6 +800,15 @@ The secrets are decrypted in a systemd user service called `sops-nix`, so other 
 }
 ```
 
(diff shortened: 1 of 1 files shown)


Mic92:
@mergify queue

PR #493, opened 2024-02-17
modules/sops/templates: Support custom files as secret templates

quentinmit:
This exposes the `file` option, which can be used with `pkgs.formats` to write additional configuration formats.

Without this option exposed, it's not possible to write configuration file formats like YAML or TOML that require a package build to generate.

### DIFF
--- a/modules/sops/templates/default.nix
+++ b/modules/sops/templates/default.nix
@@ -56,8 +56,10 @@ in {
           file = mkOption {
             type = types.path;
             default = pkgs.writeText config.name config.content;
-            visible = false;
-            readOnly = true;
+            defaultText = ''pkgs.writeText config.name config.content'';
+            description = ''
+              File used as the template.
+            '';
           };
         };
       }));


modules/sops/templates/default.nix
```diff
@@ -56,8 +56,10 @@ in {
           file = mkOption {
             type = types.path;
             default = pkgs.writeText config.name config.content;
-            visible = false;
-            readOnly = true;
+            defaultText = ''pkgs.writeText config.name config.content'';
```
Mic92, line 59:
```suggestion
            defaultText = ''pkgs.writeText config.name config.content'';
            example = "./configuration-template.conf";
```

modules/sops/templates/default.nix
```diff
@@ -56,8 +56,10 @@ in {
...
+            description = ''
+              File used as the template.
```
Mic92, line 61:
```suggestion
              File used as the template. When this value is overwritten than `sops.templates.<name>.content` is ignored.
```

quentinmit:
I've addressed your comments. PTAL and thanks!

Mic92:
@mergify queue

PR #495, opened 2024-02-18
Change sops decrypt url and update .gitignore

Janik-Haag:
Upstream changed their module url so I guess it's a good idea to updated it here.


Janik-Haag:
Forgot to change the hash, should be fixed now.

dasJ:
Sorry for the long holdup :/ Can you rebase again so we can get this in?

Janik-Haag:
> Sorry for the long holdup :/ 

no worries

> Can you rebase again so we can get this in?

done 

I think the code has some regressions? There were some test failures after the update, and i think there are more now after doing the rebase, but I don't really have the time/energy to look into it currently. Feel free to force push to here.

Mic92:
Thanks. Merged in https://github.com/Mic92/sops-nix/pull/512

Issue #496, opened 2024-02-19
Binary data embedded in yaml secret file causes template rendering error


DDoSolitary:
I have some small binary keys stored in a yaml secret file as follows:

``` yaml
some_key: !!binary <some-base64-string>
```

This format is specified in a yaml [draft doc](https://yaml.org/type/binary.html) and can be properly handled by sops.

However, when at least one template is configured in sops-nix, the substitution script tries to read my binary secret (even if it's not referenced in the template) and fails with utf-8 decoding error:

```
Setting up sops templates...
Traceback (most recent call last):
  File "/nix/store/ib527rcihx5dqmkkgwk84imrdp5cys92-substitute", line 26, in <module>
    main()
  File "/nix/store/ib527rcihx5dqmkkgwk84imrdp5cys92-substitute", line 23, in main
    print(substitute(target, subst))
          ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/ib527rcihx5dqmkkgwk84imrdp5cys92-substitute", line 15, in substitute
    content = content.replace(placeholder, f.read())
                                           ^^^^^^^^
  File "<frozen codecs>", line 322, in decode
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf9 in position 1: invalid start byte
Activation script snippet 'renderSecrets' failed (1)
```

Related: #421, #439

Issue #439, opened 2023-11-16
binary secrets are no longer interpolated


wmertens:
I have a bunch of secrets in binary format because the others don't suit. nix-sops no longer interpolates these, since https://github.com/Mic92/sops-nix/commit/7711514b8543891eea6ae84392c74a379c5010de

How about adding a secret format "text"?

Mic92:
Sounds sensible to add.

PR #497, opened 2024-02-19
Only read referenced secret on template rendering

DDoSolitary:
While we can't use non-utf8 data in template, it's not a good idea to use `config.sops.secrets.<name>.format` to determine if a secret is actually utf8 text, as yaml secrets can have embedded binary data (#496) and "binary" secrets can actually contain utf8 text (#439).

With this patch, we only read secret file when the corresponding placeholder actually occurs in template content. This way, we no longer need to try to exclude binary files from the replacement list, but rely on users to only reference text secret in templates.


Mic92:
@mergify queue

Mic92:
Thanks!

PR #509, opened 2024-03-11
Update README.md with a note for Impermanence Users

GameDungeon:
A note was added to the "Setting a user's password" part of the Read-me to help people avoid a serious pitfall. Warns Impermanence Users to link to they keyfile through there persist instead of the normal way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -591,6 +591,9 @@ $y$j9T$WFoiErKnEnMcGq0ruQK4K.$4nJAY3LBeBsZBTYSkdTOejKU6KlDmhnfUV3Ll1K/1b.
 }
 ```
 
+**Note:** If you are using Impermanence, you must set `sops.age.keyFile` to a keyfile inside your persist directory or it will not exist at boot time. 
+For example: `/nix/persist/var/lib/sops-nix/key.txt`
+
 ## Different file formats
 
 At the moment we support the following file formats: YAML, JSON, INI, dotenv and binary.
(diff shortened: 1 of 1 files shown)


byronogis:
I think there is a stright way to solve it.
Just set attr  `program.openssh.hostKeys` by self, make sure the each `hostKeys` path are storage in you persist path.
(This also can let ssh key static, not change after reboot because of data clear.)
And everything is ok, sops-nix will auto use it by [default](https://github.com/Mic92/sops-nix/blob/master/modules/sops/default.nix#L260).

Like this:

```nix
{
  program.openssh.hostKeys = [
    {
        bits = 4096;
        path = "/persist/etc/ssh/ssh_host_rsa_key";
        type = "rsa";
      }
      {
        path = "/persist/etc/ssh/ssh_host_ed25519_key";
        type = "ed25519";
      }
  ];
}
```

GameDungeon:
That might be a good additional thing to add, but it is essentially the same point except with ssh instead of key files. Moving the key files to the persist directory so they are available at boot time before impermanence has run.

byronogis:
> If you are sure this works

Yes, it is work. I build nixos and set user password with hashedPasswordFile with sops-nix.
That is [my config](https://github.com/byronogis/nix-config)

README.md
```diff
@@ -591,6 +591,9 @@ $y$j9T$WFoiErKnEnMcGq0ruQK4K.$4nJAY3LBeBsZBTYSkdTOejKU6KlDmhnfUV3Ll1K/1b.
 }
 ```
 
+**Note:** If you are using Impermanence, you must set `sops.age.keyFile` to a keyfile inside your persist directory or it will not exist at boot time. 
+For example: `/nix/persist/var/lib/sops-nix/key.txt`
```
Mic92, line 595:
```suggestion
For example: `/nix/persist/var/lib/sops-nix/key.txt`
The same is true when using ssh host keys instead of age.
```

Mic92:
@mergify queue

PR #520, opened 2024-03-25
Emit plain file when key is empty

thomaslepoix:
Renewal of #338

Fixes: #120

### DIFF
--- a/README.md
+++ b/README.md
@@ -707,6 +707,44 @@ This is how it can be included in your `configuration.nix`:
 }
 ```
 
+## Emit plain file for yaml and json formats
+
+By default, sops-nix extracts a single key from yaml and json files. If you
+need the plain file instead of extracting a specific key from the input document,
+you can set `key` to an empty string.
+
+For example, the input document `my-config.yaml` likes this:
+
+```yaml
+my-secret1: ENC[AES256_GCM,data:tkyQPQODC3g=,iv:yHliT2FJ74EtnLIeeQtGbOoqVZnF0q5HiXYMJxYx6HE=,tag:EW5LV4kG4lcENaN2HIFiow==,type:str]
+my-secret2: ENC[AES256_GCM,data:tkyQPQODC3g=,iv:yHliT2FJ74EtnLIeeQtGbOoqVZnF0q5HiXYMJxYx6HE=,tag:EW5LV4kG4lcENaN2HIFiow==,type:str]
+sops:
+    kms: []
+    gcp_kms: []
+    azure_kv: []
+    hc_vault: []
+...
+```
+
+This is how it can be included in your NixOS module:
+
+```nix
+{
+  sops.secrets.my-config = {
+    format = "yaml";
+    sopsFile = ./my-config.yaml;
+    key = "";
+  };
+}
+```
+
+Then, it will be mounted as `/run/secrets/my-config`:
+
+```yaml
+my-secret1: hello
+my-secret2: hello
+```
+
 ## Use with home manager
 
 sops-nix also provides a home-manager module.
--- a/modules/home-manager/sops.nix
+++ b/modules/home-manager/sops.nix
@@ -15,11 +15,12 @@ let
 
       key = lib.mkOption {
         type = lib.types.str;
-        default = name;
+        default = if cfg.defaultSopsKey != null then cfg.defaultSopsKey else name;
         description = ''
           Key used to lookup in the sops file.
           No tested data structures are supported right now.
           This option is ignored if format is binary.
+          "" means whole file.
         '';
       };
 
@@ -128,6 +129,16 @@ in {
       default = "yaml";
       description = ''
         Default sops format used for all secrets.
+      '';
+    };
+
+    defaultSopsKey = mkOption {
+      type = lib.types.nullOr lib.types.str;
+      default = null;
+      description = ''
+        Default key used to lookup in all secrets.
+        This option is ignored if format is binary.
+        "" means whole file.
       '';
     };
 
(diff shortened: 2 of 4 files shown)


modules/home-manager/sops.nix
```diff
@@ -124,6 +125,16 @@ in {
       '';
     };
 
+    defaultSopsKey = mkOption {
```
Mic92, line 128: Currently I don't have a better name but defaultSopsKey sounds very confusing because it's quite overloaded and we have private / public keys everywhere.

Mic92: For your usecase, do you actually need this option?
Is it not enough to set this instead?

```
{
  sops.defaultSopsFormat = "binary";
}
```

modules/home-manager/sops.nix
```diff
@@ -15,11 +15,12 @@ let
 
       key = lib.mkOption {
         type = lib.types.str;
-        default = name;
+        default = if cfg.defaultSopsKey != null then cfg.defaultSopsKey else name;
```
Mic92, line 18: Also here, can you not just set the format to "binary"?

README.md
```diff
@@ -728,6 +728,44 @@ This is how it can be included in your `configuration.nix`:
...
+    hc_vault: []
+...
+```
+
+This is how it can be included in your NixOS module:
+
+```nix
+{
+  sops.secrets.my-config = {
+    format = "yaml";
```
Mic92, line 755:
```suggestion
    format = "binary";
```

README.md
```diff
@@ -728,6 +728,44 @@ This is how it can be included in your `configuration.nix`:
...
+    sopsFile = ./my-config.yaml;
+    key = "";
```
Mic92, line 757:
```suggestion
```

thomaslepoix:
For sops, treating the secret as binary is not the same as treating it as json/yaml with an empty key (whole file), binary puts everything under a `data` key:

```sh
cat file
{
	"acl": {
		"tokens": {
			"initial_management": "911578ef-e979-4811-b607-32b1e9819b85"
		}
	}
}

sops --input-type binary -e file
{
	"data": "ENC[AES256_GCM,data:hMbIl33ak/I4SP+eJtbHwGnUn40OfagGC7AHrKk12KcLbn4M/dRjOQ+h4OwOtfp/ZSo6lyRcZKllA/DI0V6z/pL072Hs/nTkJtDnSZorCOk4THOJ84QOYVe1mcEoptX03NGN,iv:MateN0FjgcpfTcZQ7xdCIBXJW7LjchGLiy3EIo9Kt7g=,tag:7IY0D/VIclNu6oudxtl/Hw==,type:str]",
	"sops": { ... }
}

sops --input-type json -e file
{
	"acl": {
		"tokens": {
			"initial_management": "ENC[AES256_GCM,data:VChCbEmx1Mri5siKZGeIL2f2DQR7aX9YVaXtlFGnwFfkugsc,iv:0nt21nZUKaGCdviDOSCyLDvjjfajZ6UK+Mat/8LZ80I=,tag:gsZ/kavDx/YLfxGjrl5ilQ==,type:str]"
		}
	},
	"sops": { ... }
}
```

And decrypting as binary a file that was encrypted as json/yaml won't work:

```sh
sops --input-type json --output file_enc_json -e file

sops -d --output-type json file_enc_json
{
	"acl": {
		"tokens": {
			"initial_management": "911578ef-e979-4811-b607-32b1e9819b85"
		}
	}
}

sops -d --output-type binary file_enc_json
Error dumping file: error emitting binary store: no binary data found in tree

This is likely not an encrypted binary file? If not, use --output-type to select the correct output type.
```

So as detailed in the issue, using `binary` format implies the file was encrypted as such which has several drawbacks:
- It breaks sops format detection as json/yaml assumptions from the file extension becomes wrong, since sops-nix will decrypt as binary in the end. So the user has to remember this and act appropriately when encrypting secrets. Note this can go even worse when it comes to IDE plugins that brings sops support.
- It also breaks a selling feature of sops: diffable encrypted files, as everything is under the same key.
- It feels odd regarding sops that does not have such a restriction and thus generally unintuitive while learning sops-nix

---

Now, when the prefered workflow is "every secret its own file" instead of "all secrets in the same file, one secret per key", everything being json/yaml : `defaultSopsKey` just allows to be more consise than repeating `key = "";` but I agree it is really confusion-prone regarding age/pgp keys and I would like to clarify that too. To be fair this confusion is already part of sops which takes the json/yaml semantic for "key" while also dealing with cryprographic keys.

Here is an example for my current use case:

```nix
  sops = {
    age.keyFile = "/var/lib/sops-nix/key.txt";
    defaultSopsKey = "";
    defaultSopsFormat = "json";
    secrets = {
      "certs/consul-agent-ca.pem" = {
        sopsFile = self + /secrets/dev/consul-agent-ca.pem;
        format = "binary";
        owner = "consul";
      };
      "consul-encrypt.json" = {
        sopsFile = self + "/secrets/dev/consul-encrypt.json";
        owner = "consul";
        restartUnits = [ "consul-rotate-gossip-encryption-key.service" ];
      };
      "consul-acl-initial-management.json" = {
        sopsFile = self + "/secrets/dev/consul-acl-initial-management.json";
        owner = "consul";
      };
    };
  };
```

Mic92:
I see. Thanks for clarification.

BKSalman:
hello :), I'm interested in this feature, will it be merged?

thomaslepoix:
up

Mic92:
@mergify rebase

Mic92:
@mergify queue

PR #534, opened 2024-04-17
Refuse age keyfile paths that are in the nix store

the-furry-hubofeverything:
This addition would allow sops to exclude a path in the nix store on eval, specifically for `age.keyFile`.

It would defeat the purpose of a secret key if its world readable, and it's fairly simple (at least for me) to accidentally include the key file in the configuration - this is to "idiot proof" it, I guess.

I created a option type rather than an assertion, because with an assertion, the key file would be copied to the nix store on evaluation before it raises an error.

I had to rewrite the tests because they were also using key files from the store. 

Let me know if this is an good idea, or if changes are needed.


the-furry-hubofeverything:
```
       error:
       Failed assertions:
       - systemd stage 1 does not support 'boot.initrd.postDeviceCommands'. Please
         convert it to analogous systemd units in 'boot.initrd.systemd'.

           Definitions:
           - makeTest parameters
```
Okay, I'm not too familiar with `boot.initrd.systemd`, I'll try to figure it out

Mic92:
@mergify queue

lorenzleutgeb:
This change is super helpful in general. However, when checking in the key file on purpose, e.g. for a NixOS test that should use SOPS (on non-critical testing data), it gets in the way. How about adding an escape hatch `sops.age.allowInsecureKeyFileInStore`?

PR #541, opened 2024-04-18
fix symlink directory not existing

Mic92:


sedlund:
github:Mic92/sops-nix/3f167b993031a28dc137541e22ec1d8f91e0e315   works for me

PR #547, opened 2024-04-21
home-manager: minor oversight cleanup

SebTM:
Sorry my fault - small oversight ✌🏻


SebTM:
Will have a look if I find some time before the next Meetup but otherwise we could team-up on this at the next-one? :)

PR #558, opened 2024-05-05
Draft: nix-darwin module

### LINKED ISSUE #409: darwin-nix support
Just now we only support darwin via home-manager.
@Kloenk has started a work-in-progress port that could be picked up: https://github.com/Kloenk/sops-nix/commit/c0864759e5d3f5899195fd9debaeda48e08d47d8

j-baker:
Fixes #409.

The home manager module covers most cases on Darwin, but does not cover headless devices as agents don't run unless a user session exists. This is a (draft) small copy-paste job of the home-manager integration which is enough to get nix-darwin working.

I'd like feedback on:

- What tests would be useful/how this should be tested?
- What sort of code sharing should occur between this and normal Nix?
- Any documentation that'd be nice?

and honestly anything else.

### DIFF
--- a/modules/nix-darwin/sops.nix
+++ b/modules/nix-darwin/sops.nix
@@ -259,7 +259,7 @@ in {
     };
 
     # [re]load secrets on home-manager activation
-    activationScripts.sops-nix.text = let
+    system.activationScripts.sops-nix.text = let
         domain-target = "system";
       in ''
         /bin/launchctl bootout ${domain-target}/org.nixos.sops-nix && true
--- a/modules/nix-darwin/sops.nix
+++ b/modules/nix-darwin/sops.nix
@@ -259,11 +259,11 @@ in {
     };
 
     # [re]load secrets on home-manager activation
-    # activationScripts.sops-nix.text = let
-    #     domain-target = "gui/$(id -u ${config.home.username})";
-    #   in ''
-    #     /bin/launchctl bootout ${domain-target}/org.nix-community.home.sops-nix && true
-    #     /bin/launchctl bootstrap ${domain-target} ${config.home.homeDirectory}/Library/LaunchAgents/org.nix-community.home.sops-nix.plist
-    #   '';
+    activationScripts.sops-nix.text = let
+        domain-target = "system";
+      in ''
+        /bin/launchctl bootout ${domain-target}/org.nixos.sops-nix && true
+        /bin/launchctl bootstrap ${domain-target} /Library/LaunchAgents/org.nixos.sops-nix.plist
+      '';
   };
 }
--- a/modules/nix-darwin/sops.nix
+++ b/modules/nix-darwin/sops.nix
@@ -149,7 +149,7 @@ in {
 
     defaultSymlinkPath = lib.mkOption {
       type = lib.types.str;
-      default = "${config.xdg.configHome}/sops-nix/secrets";
+      default = "/var/root/sops-nix/secrets";
       description = ''
         Default place where the latest generation of decrypt secrets
         can be found.
--- a/modules/nix-darwin/sops.nix
+++ b/modules/nix-darwin/sops.nix
@@ -1,0 +1,269 @@
+{ config, lib, pkgs, ... }:
+
+let
+  cfg = config.sops;
+  sops-install-secrets = (pkgs.callPackage ../.. {}).sops-install-secrets;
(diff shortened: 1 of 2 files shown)


aymanbagabas (approved):

modules/nix-darwin/sops.nix
```diff
@@ -0,0 +1,269 @@
...
+        '';
+      };
+
+      path = lib.mkOption {
+        type = lib.types.str;
+        default = "${cfg.defaultSymlinkPath}/${name}";
+        description = ''
+          Path where secrets are symlinked to.
+          If the default is kept no other symlink is created.
+          `%r` is replaced by $XDG_RUNTIME_DIR on linux or `getconf
```
Mic92, line 32: This does not exist on macos, right?

modules/nix-darwin/sops.nix
```diff
@@ -0,0 +1,269 @@
...
+      default = "/var/root/sops-nix/secrets";
+      description = ''
+        Default place where the latest generation of decrypt secrets
+        can be found.
+      '';
+    };
+
+    defaultSecretsMountPoint = lib.mkOption {
+      type = lib.types.str;
+      default = "%r/secrets.d";
```
Mic92, line 161: Same question here.

modules/nix-darwin/sops.nix
```diff
@@ -0,0 +1,269 @@
...
+      ${pkgs.age}/bin/age-keygen -o ${cfg.age.keyFile}
+    fi
+  '' + ''
+    ${sops-install-secrets}/bin/sops-install-secrets -ignore-passwd '${manifest}'
+  '')));
+in {
+  options.sops = {
+    logFile = lib.mkOption {
+      type = lib.types.path;
+      default = "/var/log/sops.log";
```
Mic92, line 114: Do applications need to log themselves to files?

Mic92:
> I'd like feedback on:
> 
>     * What tests would be useful/how this should be tested?

Automatic tests in the CI would be great, but I don't know how this is done on macOS. I do have a macos builder in buildbot, if that helps.

> 
>     * What sort of code sharing should occur between this and normal Nix?

I don't think we need to butcher the nixos options to share them between platforms.
It's fine to just share the go code.

>     * Any documentation that'd be nice?

Yes, would be. I assume it's not that different from other platforms?

> 
> 
> and honestly anything else.

modules/nix-darwin/sops.nix
```diff
@@ -0,0 +1,269 @@
...
+      format = lib.mkOption {
+        type = lib.types.enum [ "yaml" "json" "binary" "ini" "dotenv" ];
+        default = cfg.defaultSopsFormat;
+        description = ''
+          File format used to decrypt the sops secret.
+          Binary files are written to the target file as is.
+        '';
+      };
+
+      mode = lib.mkOption {
```
Mic92, line 46: We also want owner/group here.

srid: Is there more than these two options (https://github.com/juspay/sops-nix/commit/85807169a16132dd0cb19ae2ae076480ff7dc160) that need to be done here in this PR? Because, the installed secrets are still `root` owned:

```
-sh-3.2# ls -l /var/root/sops-nix/secrets/
total 16
-r--------  1 root  wheel  13 Jun 19 02:36 example_key
-r--------  1 root  wheel   5 Jun 19 02:36 hello
```

srid:
I've tried out this PR. It sort of works (except for setting right permissions, per my comment above), but when activation fails there's no indication on the nix-darwin run side. You have to catch those silently ignored errors by tailing `/var/log/sops.log`.

Mic92:
I can now also finally properly review macOS: https://github.com/Mic92/dotfiles/tree/main/darwin

niklasravnsborg:
Hi, I'm really looking forward using the nix-darwin module, because I want to use some sensitive information in my darwin configuration and assume without this module there is no way to do that currently.

I would be very happy if you could share something on the current status of this :)

zimbatm:
Talked with @Mic92, on thing that could help is to add darwin-test

Mic92:
Not needed for merging but nice for - maintenance. Also this is the up-to-date branch: https://github.com/Mic92/sops-nix/pull/647

Mic92:
This is now merged.

PR #568, opened 2024-05-22
sops-install-secrets: add mount options for darwin

AkinoKaede:
On macOS, secretfs sometimes appears on the desktop, which is a bit troublesome for me. Mounting with the `nobrowse` option should avoid this issue.

The other mounting options were derived from already commented code. If there are any issues, please let me know.


dasJ (approved):

Mic92:
@mergify queue

PR #589, opened 2024-06-29
flake: update to 24.05

Mic92:


SuperSandro2000:
You forgot to update the flake.lock 😅 if it was that easy, I i wouldn't have created the issue in the first place.

Mic92:
> You forgot to update the flake.lock 😅 if it was that easy, I i wouldn't have created the issue in the first place.

I actually did fix it locally, but I must have gotten distracted because I did not push it.

Mic92:
@mergify queue

PR #611, opened 2024-08-31
Reuse the existing nixpkgs instance for the tests instead of re-importing it

r-vdp:


r-vdp:
Hmm, not sure why buildbot is unhappy here, this evals and builds fine locally.

Mic92:
I was testing work-in-progress buildbot-nix pull request

PR #614, opened 2024-09-05
Nix darwin module

Mic92:


niklasravnsborg:
@Mic92 I've tried to use this but for me it doesn't create the secrets in the `defaultSymlinkPath` and `defaultSecretsMountPoint`. For me it only created `age-keys.txt` and and empty `sops-nix-secretfs` file...

niklasravnsborg (changes requested):

modules/nix-darwin/sops.nix
```diff
@@ -0,0 +1,271 @@
...
+        StandardErrorPath = cfg.logFile;
+      };
+    };
+
+    # [re]load secrets on home-manager activation
+    system.activationScripts.sops-nix.text = let
+        domain-target = "system";
+      in ''
+        /bin/launchctl bootout ${domain-target}/org.nixos.sops-nix && true
+        /bin/launchctl bootstrap ${domain-target} /Library/LaunchAgents/org.nixos.sops-nix.plist
```
niklasravnsborg, line 268: This probably has to be `LaunchDaemons` instead of `LaunchAgents`...

```suggestion
        /bin/launchctl bootstrap ${domain-target} /Library/LaunchDaemons/org.nixos.sops-nix.plist
```

niklasravnsborg (changes requested):

Mic92:
> @Mic92 I've tried to use this but for me it doesn't create the secrets in the `defaultSymlinkPath` and `defaultSecretsMountPoint`. For me it only created `age-keys.txt` and and empty `sops-nix-secretfs` file...
> 
> EDIT: I think this is because my SSH-key is password protected. I found out by running: `cat /var/log/sops.log`.

password protected ssh keys are unfortunately nothing we can support, because sops runs early in boot, where we cannot prompt for passwords.

niklasravnsborg:
> password protected ssh keys are unfortunately nothing we can support, because sops runs early in boot, where we cannot prompt for passwords. I would suggest generating an age key for now.

Thank you, my solution was to use your https://github.com/Mic92/ssh-to-age to generate the age key from my encrypted SSH key and store it locally and it worked well :)

The other problem I'm now seeing, after tinkering with this module and the code a bit, is that secrets aren't available immediately after logging in for other launchd services to consume. The secrets mount is only available after [sops-install-secrets](https://github.com/Mic92/sops-nix/tree/master/pkgs/sops-install-secrets) was run. On macOS, this gets triggered by the launchd daemon, and it might take a few seconds to decrypt the secrets and mount the in-memory volume.

In my use case, I have another launchd daemon that runs a service dependent on the decrypted secrets file. But this file is not available when the service is started. I only realized sops-nix is using a volume mount after looking at the code. Before that, I assumed that the directory files would just be replaced on disk when `sops-install-secrets` is run and that they would be persisted on reboot.

For now, it feels like this might be a reason why I can't use sops-nix for my use-case. @Mic92 Do you have some thoughts on that?

niklasravnsborg:
UPDATE: In the end I used some scripting to restart my other service when the secret files are available: https://github.com/niklasravnsborg/dotfiles/commit/1afccc9c311f82d2e0cd8830141d94a158dba29a. A problem I encountered was that file watching doesn't quite work as expected for symbolic links to files inside of a volume that may not be yet mounted, when the file watcher is started. But my solution handles most cases fine.

Mic92:
@niklasravnsborg do you have a clue how I could make error reporting better in the darwin activation phase? launchd just silently fails, which is not great. I was wondering if I should not just run sops-install-secrets additionally in foreground on updates.

niklasravnsborg:
Which errors did you encounter? The `/bin/launchctl bootstrap` in the activation might not work because of the wrong path as I mentioned in a previous comment.

I guess it's fine to have it either restart the daemon or run the `sops-install-secrets` on activation but not do both.

Mic92:
I mean if the service fails, I want to see the error in the shell that runs launchd service.

niklasravnsborg:
I see, yeah this makes it harder to debug sometimes. I think seeing errors in the shell that bootstraps the service is not what launchd is made for, since launchd services run in the context of the system in the background (similarily to what systemd does on linux). On bootstraping they are just scheduled for starting and there is no immediate feedback.

So I guess we could just run the launchd service on boot but use the `sops-install-secrets` script directly in the activation. There it doesn't have to trigger the launchd service again.

I found `sudo launchctl print system/org.nixos.sops-nix` quite useful for debugging and looking at the logs.

niklasravnsborg:
@Mic92 Just an update on [my comment](https://github.com/Mic92/sops-nix/pull/614#discussion_r1749159942). This PR on nix-darwin has now been merged: https://github.com/LnL7/nix-darwin/pull/1052 and it now makes sense to update the sops-nix invocation to use `command` instead of `serviceConfig.Program`. This will ensure, that the nix store is mounted and available before trying to launch the sops-nix script.

Mic92:
Thanks for the heads up

modules/nix-darwin/sops.nix
```diff
@@ -0,0 +1,271 @@
...
+        Program = script;
+        KeepAlive = false;
+        RunAtLoad = true;
+        StandardOutPath = cfg.logFile;
+        StandardErrorPath = cfg.logFile;
+      };
+    };
+
+    # [re]load secrets on home-manager activation
+    system.activationScripts.sops-nix.text = let
```
zowoq, line 264: https://github.com/LnL7/nix-darwin/blob/master/modules/system/activation-scripts.nix

`nix-darwin` doesn't have support for arbitrary activationScripts.
```suggestion
    system.activationScripts.postActivation.text = let
```

khaneliman:
Haven't looked into it closely yet, but just figured out using this module stopped my home-manager config from being applied somehow. I was getting bootout and bootstrap errors in flake switches and noticed my home-manager changes weren't being represented in filesystem. Did a complete uninstall of nix/nix-darwin and removed nix partition to start from scratch. Saw that all my home-manager files were broken symlinks pointing to a 2201 generation. After disabling my sops-nix nix-darwin module the flake switch generated a home-manager generation and was able to get it

khaneliman:
Rebasing this branch on master with the suggested changes fixed my issue.

Mic92:
Obsolete

PR #615, opened 2024-09-05
feat(modules/home-manager): add support for configuring environment variables

marksisson:
Add support for configuring environment variables before calling `sops-install-secrets`. Introduce a new `environment` option which allows specifying environment variables. Modify systemd service and launchd agent to use the specified environment variables.

### DIFF
--- a/modules/home-manager/sops.nix
+++ b/modules/home-manager/sops.nix
@@ -96,10 +96,7 @@ let
 
   escapedAgeKeyFile = lib.escapeShellArg cfg.age.keyFile;
 
-  script = toString (pkgs.writeShellScript "sops-nix-user" ((lib.optionalString (cfg.gnupg.home != null) ''
-    export SOPS_GPG_EXEC=${pkgs.gnupg}/bin/gpg
-  '')
-    + (lib.optionalString cfg.age.generateKey ''
+  script = toString (pkgs.writeShellScript "sops-nix-user" ((lib.optionalString cfg.age.generateKey ''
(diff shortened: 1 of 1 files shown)


modules/home-manager/sops.nix
```diff
@@ -174,6 +171,18 @@ in {
...
     };
 
+    environment = lib.mkOption {
+      type = lib.types.attrsOf (lib.types.either lib.types.str lib.types.path);
+      default = {};
+      description = ''
+        Environment variables to set before calling sops-install-secrets.
+
+        The values are placed in single quotes and not escaped any further to
+        allow usage of command substitutions for more flexibility. To properly quote
```
Mic92, line 181: Is this actually true? I don't think systemd will do command substitution.

Mic92: I don't now about launchd, but it also looks to me that it should be the same.

marksisson: That may not be true. I copied the description for the environment option from the NixOS module at modules/sops/default.nix.

I also have only tested the launchd code path. It does work on macOS. I do not have access to a NixOS linux system. Hopefully someone can test the systemd code path.

Mic92:
@SebTM could you test linux?

SebTM:
Hey, sorry for the delay - I can test it but would appreciate a test-case?

Mic92:
@SebTM for now it's enough if you just make sure it doesn't break your existing usage.

Mic92:
@mergify queue

Mic92:
@mergify rebase

Issue #514, opened 2024-03-16
setting owner and group seems not to work for ids


mrVanDalo:
It seems I can't use predefined uids and gids. Here is my example.

```
  systemd.services."container@matrix-ingolf-wagner".unitConfig.ConditionPathExists = config.sops.secrets.matrix_shared_secret.path;
  sops.secrets.matrix_shared_secret.owner = toString config.ids.uids.matrix-synapse;
  sops.secrets.matrix_shared_secret.group = toString config.ids.gids.matrix-synapse;
```

Doing this I get the following error:

```
setting up /etc...
/nix/store/wizswcp7lwr4hpdrbw8jnamqa0k643wn-sops-install-secrets-0.0.1/bin/sops-install-secrets: Manifest is not valid: Failed to lookup user '224': user: unknown user 224
Activation script snippet 'setupSecrets' failed (1)
reloading user units for root...
setting up tmpfiles
warning: error(s) occurred while switching to the new configuration
```

mrVanDalo:
My current workaround is:

```
  systemd.services."container@matrix-ingolf-wagner".unitConfig.ConditionPathExists = config.sops.secrets.matrix_shared_secret.path;
  sops.secrets.matrix_shared_secret.owner = "matrix-synapse";

  users.users.matrix-synapse = {
    isSystemUser = true;
    uid = config.ids.uids.matrix-synapse;
    group = "matrix-synapse";
  };
  users.groups.matrix-synapse.gid = config.ids.gids.matrix-synapse;
```

But it's a bit strange that I have to create the users, because sops is not capable of handling ids (the documentation says it does)

Mic92:
The code should probably use https://pkg.go.dev/os/user#LookupId in addition to https://pkg.go.dev/os/user#Lookup to make this work.

clamydo:
I've stumbled over this with a different use case: I have a systemd-nspawn container via `containers` module and bind-mount a secret into it. In order to get the access right, I have to use the uid of the user _inside_ the container. Something like that:

```nix
  sops.secrets."your/secret" = {
    owner = config.containers.testvm1.config.users.users.turnserver.uid;
  };

  containers.testvm1 = let mysecret = config.sops.secrets."your/secret".path;
  in {
    autoStart = true;
    bindMounts.${mysecret}.isReadOnly = true;

    config = {
      system.stateVersion = "23.11";

      services.coturn = {
        enable = true;
        realm = "example.com";

[...]

PR #637, opened 2024-10-15
Allow to set uid and gid instead of owner and group.

munnik:
https://github.com/Mic92/sops-nix/issues/514

Allow to set uid and gid instead of owner and group. No checks will be performed when uid and gid are set.

```
sops.secrets = {
  sslCertificate = {
    sopsFile = ./secrets.yaml;
    owner = "";
    group = "";
    uid = config.containers."nginx".config.users.users."nginx".uid;
    gid = config.containers."nginx".config.users.groups."nginx".gid;
  };
  sslCertificateKey = {
    sopsFile = ./secrets.yaml;
    owner = "";
    group = "";
    uid = config.containers."nginx".config.users.users."nginx".uid;
    gid = config.containers."nginx".config.users.groups."nginx".gid;
  };
};
```

### DIFF
--- a/modules/sops/default.nix
+++ b/modules/sops/default.nix
@@ -73,18 +73,32 @@ let
         '';
       };
       owner = lib.mkOption {
-        type = lib.types.str;
-        default = "root";
+        type = with lib.types; nullOr str;
+        default = null;
         description = ''
-          User of the file.
+          User of the file. Can only be set if uid is 0.
+        '';
+      };
+      uid = lib.mkOption {
+        type = with lib.types; nullOr int;
+        default = 0;
+        description = ''
+          UID of the file, only applied when owner is null. The UID will be applied even if the corresponding user doesn't exist.
         '';
       };
       group = lib.mkOption {
-        type = lib.types.str;
-        default = users.${config.owner}.group;
+        type = with lib.types; nullOr str;
+        default = if config.owner != null then users.${config.owner}.group else null;
         defaultText = lib.literalMD "{option}`config.users.users.\${owner}.group`";
         description = ''
-          Group of the file.
+          Group of the file. Can only be set if gid is 0.
+        '';
+      };
+      gid = lib.mkOption {
+        type = with lib.types; nullOr int;
+        default = 0;
+        description = ''
+          GID of the file, only applied when group is null. The GID will be applied even if the corresponding group doesn't exist.
         '';
       };
       sopsFile = lib.mkOption {
@@ -318,6 +332,12 @@ in {
             builtins.isPath secret.sopsFile ||
             (builtins.isString secret.sopsFile && lib.hasPrefix builtins.storeDir secret.sopsFile);
           message = "'${secret.sopsFile}' is not in the Nix store. Either add it to the Nix store or set sops.validateSopsFiles to false";
+        } {
+          assertion = secret.uid != null && secret.uid != 0 -> secret.owner == null;
+          message = "In ${secret.name} exactly one of sops.owner and sops.uid must be set";
+        } {
+          assertion = secret.gid != null && secret.gid != 0 -> secret.group == null;
+          message = "In ${secret.name} exactly one of sops.group and sops.gid must be set";
         }]) cfg.secrets)
       );
 
--- a/pkgs/sops-install-secrets/main.go
+++ b/pkgs/sops-install-secrets/main.go
@@ -29,8 +29,10 @@ type secret struct {
 	Name         string     `json:"name"`
 	Key          string     `json:"key"`
 	Path         string     `json:"path"`
-	Owner        string     `json:"owner"`
-	Group        string     `json:"group"`
+	Owner        *string    `json:"owner,omitempty"`
+	UID          int        `json:"uid"`
+	Group        *string    `json:"group,omitempty"`
+	GID          int        `json:"gid"`
 	SopsFile     string     `json:"sopsFile"`
 	Format       FormatType `json:"format"`
 	Mode         string     `json:"mode"`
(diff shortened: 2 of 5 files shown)


modules/sops/default.nix
```diff
@@ -79,6 +79,13 @@ let
           User of the file.
         '';
       };
+      uid = lib.mkOption {
```
Mic92, line 82: Looks good but can you add an assertion to the assertions nixos options that make sure we are not setting both uid and owner?

modules/sops/default.nix
```diff
@@ -79,6 +79,13 @@ let
...
+        type = with lib.types; nullOr int;
+        default = -1;
```
Mic92, line 84:
```suggestion
        default = null;
```

modules/sops/default.nix
```diff
@@ -87,6 +94,13 @@ let
           Group of the file.
         '';
       };
+      gid = lib.mkOption {
+        type = with lib.types; nullOr int;
+        default = -1;
```
Mic92, line 99:
```suggestion
        default = null;
```

modules/sops/default.nix
```diff
@@ -79,6 +79,13 @@ let
           User of the file.
         '';
       };
+      uid = lib.mkOption {
+        type = with lib.types; nullOr int;
```
Mic92, line 83:
```suggestion
        type = with lib.types; nullOr types.ints.positive;
```

modules/sops/default.nix
```diff
@@ -87,6 +94,13 @@ let
           Group of the file.
         '';
       };
+      gid = lib.mkOption {
+        type = with lib.types; nullOr int;
```
Mic92, line 98:
```suggestion
        type = with lib.types; nullOr types.ints.positive;
```

Mic92:
Could you also add a simple secret example to our existing nixos test? Than we won't regress in future.

modules/sops/default.nix
Mic92: Overall I think we should prefer `uid` if it's not-null and `owner` != `root`. Otherwise it's a bit clunky if no `root` is set. The other alternative would be to switch to `uid` == 0 by default and make `owner` nullable.

modules/sops/default.nix
```diff
@@ -318,6 +332,12 @@ in {
             builtins.isPath secret.sopsFile ||
             (builtins.isString secret.sopsFile && lib.hasPrefix builtins.storeDir secret.sopsFile);
           message = "'${secret.sopsFile}' is not in the Nix store. Either add it to the Nix store or set sops.validateSopsFiles to false";
+        } {
+          assertion = secret.owner == "" && secret.uid != null || secret.owner != "" && secret.uid == null;
```
Mic92, line 336:
```suggestion
          assertion = secret.owner != "root" && secret.uid != null || secret.owner != "root" && secret.uid == null;
```

modules/sops/default.nix
```diff
@@ -318,6 +332,12 @@ in {
...
+          message = "In ${secret.name} exactly one of sops.owner and sops.uid must be set";
+        } {
+          assertion = secret.group == "" && secret.gid != null || secret.group != "" && secret.gid == null;
```
Mic92, line 339:
```suggestion
          assertion = secret.group == "root" && secret.gid != null || secret.group != "root" && secret.gid == null;
```

modules/sops/default.nix
Mic92: Also let me think about this case a bit.

pkgs/sops-install-secrets/main.go
```diff
@@ -30,7 +30,9 @@ type secret struct {
 	Key          string     `json:"key"`
 	Path         string     `json:"path"`
 	Owner        string     `json:"owner"`
+	UID          int        `json:"uid"`
```
Mic92, line 33: Is it possible to check for null || int here?

pkgs/sops-install-secrets/main.go
```diff
@@ -475,25 +477,34 @@ func (app *appContext) validateSecret(secret *secret) error {
...
-		ownerNr, err := strconv.ParseUint(owner.Uid, 10, 64)
-		if err != nil {
-			return fmt.Errorf("cannot parse uid %s: %w", owner.Uid, err)
-		}
-		secret.owner = int(ownerNr)
 
-		group, err := user.LookupGroup(secret.Group)
-		if err != nil {
-			return fmt.Errorf("failed to lookup group '%s': %w", secret.Group, err)
+		if secret.Owner == "" && secret.UID >= 0 {
```
Mic92, line 481: Because than we don't need to check if `Owner` is empty.

modules/sops/default.nix
```diff
@@ -318,6 +332,12 @@ in {
             builtins.isPath secret.sopsFile ||
             (builtins.isString secret.sopsFile && lib.hasPrefix builtins.storeDir secret.sopsFile);
           message = "'${secret.sopsFile}' is not in the Nix store. Either add it to the Nix store or set sops.validateSopsFiles to false";
+        } {
+          assertion = secret.owner != "root" && secret.uid != null || secret.owner != "root" && secret.uid == null;
```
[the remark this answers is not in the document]
Mic92:
```suggestion
          assertion = secret.owner != "root" -> secret.uid == null;
```

modules/sops/default.nix
```diff
@@ -318,6 +332,12 @@ in {
...
+          message = "In ${secret.name} exactly one of sops.owner and sops.uid must be set";
+        } {
+          assertion = secret.group == "" && secret.gid != null || secret.group != "" && secret.gid == null;
```
Mic92, line 339:
```suggestion
          assertion = secret.group != "root" -> secret.gid != null;
```

munnik: Shouldn't this be `assertion = secret.group != users.${secret.owner}.group -> secret.gid == null;` then? I'm pretty new to Nix, so I'm not sure if I understand this 100%.

pkgs/sops-install-secrets/main.go
munnik: Changed to pointers

modules/sops/default.nix
Mic92: I think you are right.

modules/sops/default.nix
munnik: I changed the code to follow your second option. I think this easier to implement.

Mic92:
@munnik `cd sops-nix; nix-build -A sops-install-secrets.tests`

Mic92:
For a single test:

```
nix-build -A sops-install-secrets.tests.age-ssh-keys
```

munnik:
> Could you also add a simple secret example to our existing nixos test? Than we won't regress in future.

Done in https://github.com/Mic92/sops-nix/pull/637/commits/a24e29e38dca7e16b88bade2c5789de396dc7273

Mic92:
@mergify queue

munnik:
Thank you! Also, for the help and feedback. Learned a lot of new Nix things. :+1:

PR #647, opened 2024-11-04
Darwin module support

ianmurphy1:
Seems like there's other PRs for this already so feel free to discard this one.

I've been trying to get this working locally on my macbook, using this branch of my own config for testing the secrets https://github.com/ianmurphy1/nixcfg/tree/darwin-sops-nix

There was a bug in it where if `config.sops.secrets` was empty then no secrets would be removed as the service would no longer be created due to the service being wrapped in a condition that there needs to be at least one secret to activate. Fixed it on darwin but this bug is present on my nixos too but I didn't touch that part as it's outside the scope of this PR.

The same bug happens with templates too which I still have to figure out so if you're ok with a bug for bug match with darwin then merge away, otherwise I'll keep working on it over the next few days as time allows to try and resolve it.

I haven't tested neededForUsers on darwin so I'll try and get that sorted too.

### DIFF
--- a/modules/nix-darwin/default.nix
+++ b/modules/nix-darwin/default.nix
@@ -1,0 +1,339 @@
+{ config, options, lib, pkgs, ... }:
+
+let
+  cfg = config.sops;
+  sops-install-secrets = cfg.package;
+  manifestFor = pkgs.callPackage ./manifest-for.nix {
+    inherit cfg;
+    inherit (pkgs) writeTextFile;
+  };
+  manifest = manifestFor "" regularSecrets {};
+
+  pathNotInStore = lib.mkOptionType {
+    name = "pathNotInStore";
+    description = "path not in the Nix store";
+    descriptionClass = "noun";
+    check = x: !lib.path.hasStorePathPrefix (/. + x);
+    merge = lib.mergeEqualOption;
+  };
+
+  regularSecrets = lib.filterAttrs (_: v: !v.neededForUsers) cfg.secrets;
+
+  withEnvironment = import ./with-environment.nix {
+    inherit cfg lib;
+  };
+  secretType = lib.types.submodule ({ config, ... }: {
+    config = {
+      sopsFile = lib.mkOptionDefault cfg.defaultSopsFile;
+      sopsFileHash = lib.mkOptionDefault (lib.optionalString cfg.validateSopsFiles "${builtins.hashFile "sha256" config.sopsFile}");
+    };
+    options = {
+      name = lib.mkOption {
+        type = lib.types.str;
+        default = config._module.args.name;
+        description = ''
+          Name of the file used in /run/secrets
+        '';
+      };
+      key = lib.mkOption {
+        type = lib.types.str;
+        default = config._module.args.name;
+        description = ''
+          Key used to lookup in the sops file.
+          No tested data structures are supported right now.
+          This option is ignored if format is binary.
+        '';
+      };
+      path = lib.mkOption {
+        type = lib.types.str;
+        default = if config.neededForUsers then "/run/secrets-for-users/${config.name}" else "/run/secrets/${config.name}";
+        defaultText = "/run/secrets-for-users/$name when neededForUsers is set, /run/secrets/$name when otherwise.";
+        description = ''
+          Path where secrets are symlinked to.
+          If the default is kept no symlink is created.
+        '';
+      };
+      format = lib.mkOption {
+        type = lib.types.enum ["yaml" "json" "binary" "dotenv" "ini"];
+        default = cfg.defaultSopsFormat;
+        description = ''
+          File format used to decrypt the sops secret.
+          Binary files are written to the target file as is.
+        '';
+      };
+      mode = lib.mkOption {
+        type = lib.types.str;
+        default = "0400";
+        description = ''
+          Permissions mode of the in octal.
+        '';
+      };
+      owner = lib.mkOption {
+        type = with lib.types; nullOr str;
+        default = "root";
+        description = ''
+          User of the file. Can only be set if uid is 0.
+        '';
+      };
+      uid = lib.mkOption {
+        type = with lib.types; nullOr int;
+        default = 0;
+        description = ''
+          UID of the file, only applied when owner is null. The UID will be applied even if the corresponding user doesn't exist.
+        '';
+      };
+      group = lib.mkOption {
+        type = with lib.types; nullOr str;
+        default = "staff";
+        defaultText = "staff";
+        description = ''
+          Group of the file. Can only be set if gid is 0.
+        '';
+      };
+      gid = lib.mkOption {
+        type = with lib.types; nullOr int;
+        default = 0;
+        description = ''
+          GID of the file, only applied when group is null. The GID will be applied even if the corresponding group doesn't exist.
+        '';
+      };
+      sopsFile = lib.mkOption {
+        type = lib.types.path;
+        defaultText = lib.literalExpression "\${config.sops.defaultSopsFile}";
+        description = ''
+          Sops file the secret is loaded from.
+        '';
+      };
+      sopsFileHash = lib.mkOption {
+        type = lib.types.str;
+        readOnly = true;
+        description = ''
+          Hash of the sops file.
+        '';
+      };
+      neededForUsers = lib.mkOption {
+        type = lib.types.bool;
+        default = false;
+        description = ''
+           **Warning** This option doesn't have any effect on macOS, as nix-darwin cannot manage user passwords on macOS.
+          This can be used to retrieve user's passwords from sops-nix.
+          Setting this option moves the secret to /run/secrets-for-users and disallows setting owner and group to anything else than root.
+        '';
+      };
+    };
+  });
+
+  darwinSSHKeys = [{
+    type = "rsa";
+    path = "/etc/ssh/ssh_host_rsa_key";
+  } {
+    type = "ed25519";
+    path = "/etc/ssh/ssh_host_ed25519_key";
+  }];
+
+  escapedKeyFile = lib.escapeShellArg cfg.age.keyFile;
+  # Skip ssh keys deployed with sops to avoid a catch 22
+  defaultImportKeys = algo:
+    map (e: e.path) (lib.filter (e: e.type == algo && !(lib.hasPrefix "/run/secrets" e.path)) darwinSSHKeys);
+
+  installScript = ''
+    ${if cfg.age.generateKey then ''
+    if [[ ! -f ${escapedKeyFile} ]]; then
+      echo generating machine-specific age key...
+      mkdir -p $(dirname ${escapedKeyFile})
+      # age-keygen sets 0600 by default, no need to chmod.
+      ${pkgs.age}/bin/age-keygen -o ${escapedKeyFile}
+    fi
+    '' else ""}
+    echo "Setting up secrets..."
+    ${withEnvironment "${sops-install-secrets}/bin/sops-install-secrets ${manifest}"}
+  '';
+
+in {
+  options.sops = {
+    secrets = lib.mkOption {
+      type = lib.types.attrsOf secretType;
+      default = {};
+      description = ''
+        Path where the latest secrets are mounted to.
+      '';
+    };
+
+    defaultSopsFile = lib.mkOption {
+      type = lib.types.path;
+      description = ''
+        Default sops file used for all secrets.
+      '';
+    };
+
+    defaultSopsFormat = lib.mkOption {
+      type = lib.types.str;
+      default = "yaml";
+      description = ''
+        Default sops format used for all secrets.
+      '';
+    };
+
+    validateSopsFiles = lib.mkOption {
+      type = lib.types.bool;
+      default = true;
+      description = ''
+        Check all sops files at evaluation time.
+        This requires sops files to be added to the nix store.
+      '';
+    };
+
+    keepGenerations = lib.mkOption {
+      type = lib.types.ints.unsigned;
+      default = 1;
+      description = ''
+        Number of secrets generations to keep. Setting this to 0 disables pruning.
+      '';
+    };
+
+    log = lib.mkOption {
+      type = lib.types.listOf (lib.types.enum [ "keyImport" "secretChanges" ]);
+      default = [ "keyImport" "secretChanges" ];
+      description = "What to log";
+    };
+
+    environment = lib.mkOption {
+      type = lib.types.attrsOf (lib.types.either lib.types.str lib.types.path);
+      default = {};
+      description = ''
+        Environment variables to set before calling sops-install-secrets.
+
+        The values are placed in single quotes and not escaped any further to
+        allow usage of command substitutions for more flexibility. To properly quote
+        strings with quotes use lib.escapeShellArg.
+
+        This will be evaluated twice when using secrets that use neededForUsers but
+        in a subshell each time so the environment variables don't collide.
+      '';
+    };
+
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = (pkgs.callPackage ../.. {}).sops-install-secrets;
+      defaultText = lib.literalExpression "(pkgs.callPackage ../.. {}).sops-install-secrets";
+      description = ''
+        sops-install-secrets package to use.
+      '';
+    };
+
+    validationPackage = lib.mkOption {
+      type = lib.types.package;
+      default =
+        if pkgs.stdenv.buildPlatform == pkgs.stdenv.hostPlatform
+          then sops-install-secrets
+          else (pkgs.pkgsBuildHost.callPackage ../.. {}).sops-install-secrets;
+      defaultText = lib.literalExpression "config.sops.package";
+
+      description = ''
+        sops-install-secrets package to use when validating configuration.
+
+        Defaults to sops.package if building natively, and a native version of sops-install-secrets if cross compiling.
+      '';
+    };
+
+    age = {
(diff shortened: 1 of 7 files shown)


modules/nix-darwin/templates/default.nix
```diff
@@ -0,0 +1,121 @@
...
+      launchd.daemons.sops-nix-templates = {
+        script = renderScript;
+        serviceConfig = {
+          KeepAlive = false;
+          RunAtLoad = true;
+        };
+      };
+      system.activationScripts = {
+        sops-nix-templates.text = ''
+          launchctl unload /Library/LaunchDaemons/org.nixos.sops-nix-templates.plist
```
Mic92, line 115: Do you have maybe any idea how we show errors in the action phase from sops nix?

ianmurphy1: I do not, any modules i see logs for when activating are because they use activation scripts vs launchd itself.

Home manager for instance I see this being logged: https://github.com/nix-community/home-manager/blob/8f6ca7855d409aeebe2a582c6fd6b6a8d0bf5661/nix-darwin/default.nix#L17

launchctl is bad from what I can see, and the mac version is even more stripped down. Launchd may not be the answer and instead an activation script is instead the way to go with it.

I'll change the install secrets from launchd to script based and see how it goes as the lack of insight into launchd is annoying

modules/nix-darwin/default.nix
```diff
@@ -0,0 +1,374 @@
...
+      sops.environment.SOPS_GPG_EXEC = lib.mkIf (cfg.gnupg.home != null || cfg.gnupg.sshKeyPaths != []) (lib.mkDefault "${pkgs.gnupg}/bin/gpg");
+
+      system.build.sops-nix-manifest = manifest;
+      system.activationScripts = {
+        postActivation.text = ''
+          echo "setting up secrets..."
+          ${sops-install-secrets}/bin/sops-install-secrets ${manifest}
+        '';
+      };
+
```
ianmurphy1, line 347: @Mic92 This results in the following

```
setting up secrets...
sops-install-secrets: Imported /etc/ssh/ssh_host_rsa_key as GPG key with fingerprint 500af0f7aa68b24d0f8c280a3a5caf201b8873ef
sops-install-secrets: Imported /etc/ssh/ssh_host_ed25519_key as age key with fingerprint age1jzf4tszg7ny9nzs54k4vkzfl3rqhp5aczwew47enhezauhgsj5tq05a29d
```

Being outputted during a `darwin-rebuild`

Mic92: Ah. yeah. I think I prefer it this way.

modules/nix-darwin/templates/default.nix
```diff
@@ -0,0 +1,121 @@
...
+            "${toString config.sops.placeholder.${name}} ${
+              config.sops.secrets.${name}.path
+            }") (attrNames config.sops.secrets));
+      in ''
+        mkdir -p "${dirOf tpl.path}"
+        (umask 077; ${substitute} ${tpl.file} ${subst-pairs} > ${tpl.path})
+        chmod "${tpl.mode}" "${tpl.path}"
+        chown "${tpl.owner}:${tpl.group}" "${tpl.path}"
+      '') (attrNames config.sops.templates)}
+  '';
```
Mic92, line 27: We can remove the python script now that templates have been added to the go code: https://github.com/Mic92/sops-nix/pull/649

modules/nix-darwin/templates/default.nix
```diff
@@ -0,0 +1,121 @@
+{ config, pkgs, lib, options, ... }:
+with lib;
+with lib.types;
+with builtins;
```
Mic92, line 4: We also stripped those in the nixos version now:

```suggestion
```

modules/nix-darwin/templates/subs.py
```diff
@@ -0,0 +1,25 @@
...
+    return content
+
+
+def main() -> None:
+    target = argv[1]
+    subst = argv[2]
+    print(substitute(target, subst))
+
+
+main()
```
Mic92, line 25:
```suggestion
```

Mic92:
If this template changes are done. I think this good for another test run.

ianmurphy1:
Brought in latest code from sops-nix to change to golang templates instead of python. Messed it up by the looks of it as it's showing a lot of file changes.

No changes other than the merge and then the change to the templates to match current sops implementation. I'll have a chance later to test it out locally and see how the golang templates are working out.

ianmurphy1:
Rebased so now there's just my changes in the PR.

ianmurphy1:
Main thing of note from latest commit is using launchd to make secrets populate at boot time.

There's only a subset of activation scripts ran during boot: [activate-system](https://github.com/LnL7/nix-darwin/blob/master/modules/services/activate-system/default.nix) meaning secrets weren't there on initial boot, it needed a run of `darwin-rebuild` to populate them. Activation scripts are used thereafter.

Keyruu:
@Mic92 told me to test this, so I did and it works. Awesome work!

khaneliman:
Been using this Pr myself and seen no issues, as well.

ianmurphy1:
I've squashed this into a single commit to avoid clobbering the commit history.

There's no support for setting user passwords in darwin from what I can see.

Might be worth keeping in for feature parity but could just not be imported until it is in nix-darwin and then allow it to be imported.

pkgs/sops-install-secrets/darwin.go
```diff
@@ -71,21 +71,21 @@ func MountSecretFs(mountpoint string, keysGID int, _useTmpfs bool, userMode bool
 	size := mb * 1024 * 1024 / 512 // size in sectors a 512 bytes
 	cmd := exec.Command("hdiutil", "attach", "-nomount", fmt.Sprintf("ram://%d", int(size)))
 	out, err := cmd.Output() // /dev/diskN
-	log.Printf("%q\n", string(out))
+	//log.Printf("%q\n", string(out))
```
Mic92, line 74: If we no longer want these outputs, we should rather delete them.

modules/nix-darwin/default.nix
```diff
@@ -0,0 +1,366 @@
...
+  } {
+    type = "rsa";
+    path = "/etc/ssh/ssh_host_rsa_key";
+  } {
+    type = "ecdsa";
+    path = "/etc/ssh/ssh_host_ecdsa_key";
+  } {
+    type = "ed25519";
+    path = "/etc/ssh/ssh_host_ed25519_key";
+  }];
```
Mic92, line 138: These formats are not supported:

```suggestion
  darwinSSHKeys = [{
    type = "rsa";
    path = "/etc/ssh/ssh_host_rsa_key";
  } {
    type = "ed25519";
    path = "/etc/ssh/ssh_host_ed25519_key";
  }];
```

modules/nix-darwin/default.nix
```diff
@@ -0,0 +1,366 @@
...
+
+        ```
+        swapDevices = [{
+          device = "/dev/sdXY";
+          randomEncryption.enable = true;
+        }];
+        ```
+      '';
+    };
+
```
Mic92, line 264: There is no tmpfs in macOS, only ramfs.
```suggestion
```

modules/nix-darwin/manifest-for.nix
```diff
@@ -0,0 +1,29 @@
...
+    secrets = builtins.attrValues secrets;
+    # Does this need to be configurable?
+    secretsMountPoint = "/run/secrets.d";
+    symlinkPath = "/run/secrets";
+    keepGenerations = cfg.keepGenerations;
+    gnupgHome = cfg.gnupg.home;
+    sshKeyPaths = cfg.gnupg.sshKeyPaths;
+    ageKeyFile = cfg.age.keyFile;
+    ageSshKeyPaths = cfg.age.sshKeyPaths;
+    useTmpfs = cfg.useTmpfs;
```
Mic92, line 17:
```suggestion
    useTmpfs = false;
```

modules/nix-darwin/default.nix
```diff
@@ -0,0 +1,366 @@
...
+        When using a swap file or device, consider enabling swap encryption by setting the `randomEncryption.enable` option
+
+        ```
+        swapDevices = [{
+          device = "/dev/sdXY";
+          randomEncryption.enable = true;
+        }];
+        ```
+      '';
+    };
```
Mic92, line 263:
```suggestion
```

modules/nix-darwin/default.nix
```diff
@@ -0,0 +1,366 @@
...
+        '';
+      };
+      sopsFileHash = lib.mkOption {
+        type = lib.types.str;
+        readOnly = true;
+        description = ''
+          Hash of the sops file.
+        '';
+      };
+      neededForUsers = lib.mkOption {
```
Mic92, line 114: How far away from managing user passwords are we actually in macOS? Is this actually something that can be done from a CLI without user interaction?

modules/nix-darwin/secrets-for-users/default.nix
```diff
@@ -0,0 +1,43 @@
...
+  assertions = [{
+    assertion = (lib.filterAttrs (_: v: (v.uid != 0 && v.owner != "root") || (v.gid != 0 && v.group != "root")) secretsForUsers) == { };
+    message = "neededForUsers cannot be used for secrets that are not root-owned";
+  }];
+
+  system.activationScripts =  {
+    postActivation.text = lib.mkAfter installScript;
+  };
+
+  launchd.daemons.sops-install-secrets-for-users = {
```
Mic92, line 33:
```suggestion
  launchd.daemons.sops-install-secrets-for-users = lib.mkIf (secretsForUsers != []) {
```

modules/nix-darwin/default.nix
```diff
@@ -0,0 +1,366 @@
...
+          This option must be explicitly unset if <literal>config.sops.gnupg.home</literal> is set.
+        '';
+      };
+    };
+  };
+  imports = [
+    ./templates
+    ./secrets-for-users
+    (lib.mkRenamedOptionModule [ "sops" "gnupgHome" ] [ "sops" "gnupg" "home" ])
+    (lib.mkRenamedOptionModule [ "sops" "sshKeyPaths" ] [ "sops" "gnupg" "sshKeyPaths" ])
```
Mic92, line 320: We never had these options in the first place:

```suggestion
```

modules/nix-darwin/secrets-for-users/default.nix
```diff
@@ -0,0 +1,43 @@
...
+  '';
+in
+{
+
+  assertions = [{
+    assertion = (lib.filterAttrs (_: v: (v.uid != 0 && v.owner != "root") || (v.gid != 0 && v.group != "root")) secretsForUsers) == { };
+    message = "neededForUsers cannot be used for secrets that are not root-owned";
+  }];
+
+  system.activationScripts =  {
```
Mic92, line 29:
```suggestion
  system.activationScripts = lib.mkIf (secretsForUsers != []) {
```

modules/nix-darwin/default.nix
```diff
@@ -0,0 +1,366 @@
...
+        readOnly = true;
+        description = ''
+          Hash of the sops file.
+        '';
+      };
+      neededForUsers = lib.mkOption {
+        type = lib.types.bool;
+        default = false;
+        description = ''
+          Enabling this option causes the secret to be decrypted before users and groups are created.
```
Mic92, line 118:
```suggestion
          **Warning** This option doesn't have any effect on macOS, as nix-darwin cannot manage user passwords on macOS.
          Enabling this option causes the secret to be decrypted before users and groups are created.
```

modules/nix-darwin/default.nix
ianmurphy1: It's possible: [mac user](https://apple.stackexchange.com/questions/226073/how-do-i-create-user-accounts-from-the-terminal-in-mac-os-x-10-11/292859#292859)

[`nix-darwin`](https://github.com/LnL7/nix-darwin/blob/master/modules/users/default.nix) doesn't implement it though.

modules/nix-darwin/secrets-for-users/default.nix
```diff
@@ -0,0 +1,43 @@
...
+  };
+
+  launchd.daemons.sops-install-secrets-for-users = {
+    command = installScript;
+    serviceConfig = {
+      RunAtLoad = true;
+      KeepAlive = false;
+    };
+  };
+  sops.environment.SOPS_GPG_EXEC = lib.mkIf (cfg.gnupg.home != null || cfg.gnupg.sshKeyPaths != []) (lib.mkDefault "${pkgs.gnupg}/bin/gpg");
```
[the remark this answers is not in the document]
ianmurphy1: Was testing something locally, missed it in the cleanup, removed

ianmurphy1:
Pretty sure I've gotten all suggested changes in this latest push

dudeofawesome:
I've been trying out this PR on one of my machines, and I'm seeing some issues with templates. 

I've got a simple config here defining 2 secrets and merging them into a single config file for use with a gitlab-runner: 

```nix
{ config, ... }:
let
  server_url = "hosts/lab-mac-mini/gitlab-runner/server_url";
  auth_token = "hosts/lab-mac-mini/gitlab-runner/auth_token";
in
{
  sops.secrets.${server_url}.sopsFile = ./secrets.yaml;
  sops.secrets.${auth_token}.sopsFile = ./secrets.yaml;

  sops.templates."gitlab-runner-registration.env".content = ''
    CI_SERVER_URL=${config.sops.placeholder.${server_url}}
    CI_SERVER_TOKEN=${config.sops.placeholder.${auth_token}}
  '';
}
```

The 1st time I try to `darwin-rebuild switch` to the flake including the `sops.templates.…`, I get this error at the end of the logs:

```
...
Activating setDarwinDefaults
Activating setupLaunchAgents
Setting up secrets...
sops-install-secrets: Imported /etc/ssh/ssh_host_rsa_key as GPG key with fingerprint 50fdf92d0f8e51cbe5546afb7b156a311e835379
sops-install-secrets: Imported /etc/ssh/ssh_host_ed25519_key as age key with fingerprint age1yrjflhg64fs9lt36qx8k96rjfyqvghrl4t5x6uwscg6umw4ev3qq4vpaku
Setting up secrets for users
sops-install-secrets: Imported /etc/ssh/ssh_host_rsa_key as GPG key with fingerprint 50fdf92d0f8e51cbe5546afb7b156a311e835379
sops-install-secrets: Imported /etc/ssh/ssh_host_ed25519_key as age key with fingerprint age1yrjflhg64fs9lt36qx8k96rjfyqvghrl4t5x6uwscg6umw4ev3qq4vpaku
adding rendered secret: gitlab-runner-registration.env
/nix/store/j4mb6a2d7q78mzza0lqprbq0h4jsaigk-sops-install-secrets-0.0.1/bin/sops-install-secrets: failed to prepare symlinks to secret store: failed to symlink template '/run/secrets/rendered/gitlab-runner-registration.env': cannot chown symlink '/run/secrets/rendered/gitlab-runner-registration.env': failed to open /run/secrets/rendered/gitlab-runner-registration.env: too many levels of symbolic links
```

If I try to the `darwin-rebuild switch` again, I get a different error:

```
...
Activating setDarwinDefaults
Activating setupLaunchAgents
Setting up secrets...
sops-install-secrets: Imported /etc/ssh/ssh_host_rsa_key as GPG key with fingerprint 50fdf92d0f8e51cbe5546afb7b156a311e835379
sops-install-secrets: Imported /etc/ssh/ssh_host_ed25519_key as age key with fingerprint age1yrjflhg64fs9lt36qx8k96rjfyqvghrl4t5x6uwscg6umw4ev3qq4vpaku
/nix/store/j4mb6a2d7q78mzza0lqprbq0h4jsaigk-sops-install-secrets-0.0.1/bin/sops-install-secrets: cannot request units to restart: lstat /private/var/run/secrets-for-users.d/7/rendered: no such file or directory
```

---

I also noticed that if `/run/secrets/` already exists, but was not created by this project, I'll get this error message:

```
...
Activating setupLaunchAgents
Setting up secrets...
sops-install-secrets: Imported /etc/ssh/ssh_host_rsa_key as GPG key with fingerprint 8e3f62a2c899a48b38727f8068a031535b0a32b2
sops-install-secrets: Imported /etc/ssh/ssh_host_ed25519_key as age key with fingerprint age1rqjddanhdhc3gvj9mj9ez5cm492uqtcnj9elljpp4uy4hx4p3sgqyyfxa6
/nix/store/j4mb6a2d7q78mzza0lqprbq0h4jsaigk-sops-install-secrets-0.0.1/bin/sops-install-secrets: failed to prepare new secrets directory: cannot access /run/secrets: readlink /run/secrets: invalid argument
```

I was able to solve this issue by simply `rm -rf`ing the pre-existing dir.

---

Other than those 2 issues though, it looks like the secrets themselves are working great!

Mic92:
@mergify queue

Mic92:
@ianmurphy1 @dudeofawesome Some documentation paragraph in the README would be nice. i.e. how to use this with flakes.

ianmurphy1:
@Mic92 I'll try and get something in Monday or Tuesday unless I get a spare moment over the weekend to do it.

@dudeofawesome Not sure about this as it's not something I've encountered as the only secrets I have are managed by sops-nix so there's no other tool using the `/run/secrets` dir to get in the way. Like above I'm not likely to get to this until early next week to try and replicate locally. Have you tried the same thing on NixOS and gotten similar errors or is it darwin specific?

dudeofawesome:
I’d been messing around with manually placing secrets in that directory prior to this PR, which is why it existed at all. I don’t intend to do that going forward.

Issue #652, opened 2024-11-07
Bug: activation messages get confused about rendered templates


jfly:
With a configuration like this:

```nix
sops.templates.test_template.content = ''
  this is a template
'';
```

Deploy once:

```
$ /run/current-system/bin/switch-to-configuration
```

Note the message here about removing `rendered/test_template`:

```
$ /run/current-system/bin/switch-to-configuration dry-activate
machine # would activate the configuration...
would remove secret: rendered/test_template
```

We broke this in https://github.com/Mic92/sops-nix/commit/bb7d6362119b792aeebe1926da1b5ebe37fed0c3#diff-de10bfc5a14f955e99213bfd939fcfbc980cf613abdb34228478bf488651fc24R46 because we changed the default path for templates from `/run/secrets-rendered/` to `/run/secrets/rendered/`, which then confuses [this diffing code](https://github.com/Mic92/sops-nix/blob/c5ae1e214ff935f2d3593187a131becb289ea639/pkgs/sops-install-secrets/main.go#L955-L971) which recurses through the entire `/run/secrets` directory.

PR #654, opened 2024-11-07
Improve activation messages about rendered templates

jfly:
This PR contains 2 commits, I suggest reading them independently:

1. The first commit adds a test that demonstrates <https://github.com/Mic92/sops-nix/issues/652>. It doesn't actually fix the bug.
2. The second commit fixes https://github.com/Mic92/sops-nix/issues/652


Mic92:
@mergify queue

Issue #653, opened 2024-11-07
Bug: templates.path option no longer works


jfly:
In https://github.com/Mic92/sops-nix/commit/aa5caa129bd96b7883fa5164dae7d91279ecb9d2#diff-de10bfc5a14f955e99213bfd939fcfbc980cf613abdb34228478bf488651fc24L25, we removed the code that would render templates to the chosen location.

Now they *only* get written to `/run/secrets/rendered/<template-name>`.

Instead of directly writing to the location the user chose (as the old code did), I believe we should instead produce a symlink to rendered template. This would give us a mechanism for cleaning up those old files when we're done with them (although that is unfortunately currently broken. see https://github.com/Mic92/sops-nix/issues/574 for details)

PR #655, opened 2024-11-07
fix: create `template.path` symlink

jfly:
This fixes https://github.com/Mic92/sops-nix/issues/653.

Note: `main.go` has been slowly accumulating shared logic between vanilla "secrets" and "templates". It feels to me like we could DRY up some of the logic in here by creating some shared "interface" that they both implement. I opted not to try to tackle that here, though.

### DIFF
--- a/pkgs/sops-install-secrets/main.go
+++ b/pkgs/sops-install-secrets/main.go
@@ -172,51 +172,51 @@ func readManifest(path string) (*manifest, error) {
 	return &m, nil
 }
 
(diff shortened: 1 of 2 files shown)


Mic92:
@mergify queue

DrymarchonShaun:
This might need to be reverted, it seems to be causing `/run/secrets/` to not be symlinked to `/run/secrets.d/1/*` in certain situations, (see #659).

jfly:
I should be able to look into this in a few hours. @DrymarchonShaun, I doubt it's this PR that introduced your issue. I'd be suspicious of https://github.com/Mic92/sops-nix/pull/649.

PR #657, opened 2024-11-09
home-manager: Add support for Split GPG on Qubes OS

kalbasit:
I'm using Qubes OS and I've configured my Qubes to use [Split GPG](https://www.qubes-os.org/doc/split-gpg). This change adds support for sops-nix to run `/usr/bin/qubes-gpg-client-wrapper` instead of the typical gpg binary. The environment also needs the name of the Qube that has the GPG keys in the environment variable named `QUBES_GPG_DOMAIN`.

I've added myself a todo to consider creating a Nix package for the `/usr/bin/qubes-gpg-client-wrapper` binary available in [this repository](https://github.com/QubesOS/qubes-app-linux-split-gpg). I have not done that right now because I'm new to Qubes OS and I'm not sure what happens if the client version is different than the server running in the vault.

### DIFF
--- a/modules/home-manager/sops.nix
+++ b/modules/home-manager/sops.nix
@@ -231,6 +231,19 @@ in {
         '';
       };
 
+      qubes-split-gpg = {
+        enable = lib.mkEnableOption "Enable support for Qubes Split GPG feature: https://www.qubes-os.org/doc/split-gpg";
+
+        domain = lib.mkOption {
+          type = lib.types.nullOr lib.types.str;
+          default = null;
+          example = "vault-gpg";
+          description = ''
+            It tells Qubes OS which secure Qube holds your GPG keys for isolated cryptographic operations.
+          '';
+        };
+      };
+
       sshKeyPaths = lib.mkOption {
         type = lib.types.listOf lib.types.path;
         default = [];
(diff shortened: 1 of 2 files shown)


modules/home-manager/sops.nix
```diff
@@ -244,24 +257,50 @@ in {
...
+            (builtins.isString secret.sopsFile && lib.hasPrefix builtins.storeDir secret.sopsFile);
         message = "'${secret.sopsFile}' is not in the Nix store. Either add it to the Nix store or set sops.validateSopsFiles to false";
       }]) cfg.secrets)
     );
 
-    sops.environment.SOPS_GPG_EXEC = lib.mkIf (cfg.gnupg.home != null || cfg.gnupg.sshKeyPaths != []) (lib.mkDefault "${pkgs.gnupg}/bin/gpg");
+    home.sessionVariables = lib.mkIf cfg.gnupg.qubes-split-gpg.enable {
+      # TODO: Add this package to nixpkgs and use it from the store
+      # https://github.com/QubesOS/qubes-app-linux-split-gpg
+      SOPS_GPG_EXEC = "/usr/bin/qubes-gpg-client-wrapper";
```
Mic92, line 293: In case someone ever packages it, wouldn't it be better to use a PATH lookup?

```suggestion
      SOPS_GPG_EXEC = "qubes-gpg-client-wrapper";
```

kalbasit: I think that's a fair callout, Let me merge it and try it to make sure it still works.

kalbasit: Yep it works perfectly!

Issue #659, opened 2024-11-10
I'm running into a super strange issue with /run/secrets not working


dkowis:
I suspect it may have something to do with recent changes that mention moving this directory around:
-  https://github.com/Mic92/sops-nix/issues/652

My problem is that when the system reboots, I don't get a proper `/run/secrets` directory that is a symlink to the right location.
I have  a directory `/run/secrets/rendered` instead of it being a symlink to `/run/secrets.d/1` with a rendered directory in it?

This host has an early secrets-for-users in a template, that probably doesn't need to be in a template, but I think that's what's happening.

I get this error on activation:
```
[root@sarafan:/etc/nixos]# nixos-rebuild switch
building the system configuration...
activating the configuration...
sops-install-secrets: Imported /etc/ssh/ssh_host_rsa_key as GPG key with fingerprint 8205c05ea62380d8950426c019a42470335483ff
sops-install-secrets: Imported /etc/ssh/ssh_host_ed25519_key as age key with fingerprint age18g8r7drww22ulsw3728n4nz9g5kcn0tvvvsw0dql89kkufttxe3s72j6fj
setting up /etc...
sops-install-secrets: Imported /etc/ssh/ssh_host_rsa_key as GPG key with fingerprint 8205c05ea62380d8950426c019a42470335483ff
sops-install-secrets: Imported /etc/ssh/ssh_host_ed25519_key as age key with fingerprint age18g8r7drww22ulsw3728n4nz9g5kcn0tvvvsw0dql89kkufttxe3s72j6fj
/nix/store/aax7zp4j2pi18mnig3xmzb64j9axvmww-sops-install-secrets-0.0.1/bin/sops-install-secrets: failed to prepare new secrets directory: cannot access /run/secrets: readlink /run/secrets: invalid argument
Activation script snippet 'setupSecrets' failed (1)
reloading user units for nixboi...
restarting sysinit-reactivation.target
warning: the following units failed: mysql.service, postgresql.service
```

And since my ssl key is sops'd it doesn't show up. So far, the only solution I've had is to go back in time generation wise, built on 9 Nov 2024.

dkowis:
Yep, confirmed. If I change that template to not be neededForUsers, everything shows up fine.

I think it's an order of operations with the changes for the template directories rendering.

DrymarchonShaun:
I'm also seeing the same error, and I don't have any templates using neededForUsers. It seems the problem appeared from #655.

jfly:
[...]

PR #660, opened 2024-11-11
Demonstration of how system activation fails if the template path is not "external"

jfly:
This demonstrates one way of reproducing
https://github.com/Mic92/sops-nix/issues/659. There may be others: I think you could achieve the same thing with a regular secret rather than a template.

Here's how this fails:

    $ nix build .#checks.x86_64-linux.template-path-not-external -L
    ...
    vm-test-run-sops-template-path-not-external> machine: must succeed: /run/current-system/bin/switch-to-configuration test
    vm-test-run-sops-template-path-not-external> machine # [   11.028130] nixos[949]: switching to system configuration /nix/store/0jbh5rys5x1br66s4n777jq8ny2bq9nb-nixos-system-machine-test
    vm-test-run-sops-template-path-not-external> machine # [   11.029653] systemd[1]: Stopped target Local File Systems.
    vm-test-run-sops-template-path-not-external> machine # [   11.031619] systemd[1]: Stopped target Remote File Systems.
    vm-test-run-sops-template-path-not-external> machine # activating the configuration...
    vm-test-run-sops-template-path-not-external> machine # /nix/store/8788mz0w701ff12zpsgnbm7nzsfy290k-sops-install-secrets-0.0.1/bin/sops-install-secrets: failed to prepare new secrets directory: cannot access /run/secrets: readlink /run/secrets: invalid argument
    vm-test-run-sops-template-path-not-external> machine # Failed to run activate script
    ...

The only fix I can think of is to check if the chosen path is inside of `/run/secrets` and reject it if it is (of course allowing for the default value for these paths, which actually *is* inside of `/run/secrets`).

### DIFF
--- a/pkgs/sops-install-secrets/nixos-test.nix
+++ b/pkgs/sops-install-secrets/nixos-test.nix
@@ -331,6 +331,43 @@ in {
     '';
   };
 
+  template-path-not-external = testers.runNixOSTest {
+    name = "sops-template-path-not-external";
+    nodes.machine = {config, ...}: {
+      imports = [ ../../modules/sops ];
+      sops = {
+        age.keyFile = "/run/age-keys.txt";
+        defaultSopsFile = ./test-assets/secrets.yaml;
+        secrets.test_key = { };
+      };
+
+      # Must run before sops sets up keys.
+      boot.initrd.postDeviceCommands = ''
+        cp -r ${./test-assets/age-keys.txt} /run/age-keys.txt
+        chmod -R 700 /run/age-keys.txt
+      '';
+
+      sops.templates.test_template = {
+        content = ''
+          Test value: ${config.sops.placeholder.test_key}
+        '';
+        # This path is inside of `/run/secrets`, which isn't allowed.
+        path = "/run/secrets/foo-bar";
+      };
+
+      system.switch.enable = true;
+    };
+
+    testScript = ''
+      start_all()
+      machine.wait_for_unit("multi-user.target")
+      # TODO: is there some way to assert that system activation worked during
+      # the initial boot? We don't actually have to try activating again
+      # here...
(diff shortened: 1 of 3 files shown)


pkgs/sops-install-secrets/nixos-test.nix
```diff
@@ -331,6 +331,43 @@ in {
...
+        cp -r ${./test-assets/age-keys.txt} /run/age-keys.txt
+        chmod -R 700 /run/age-keys.txt
+      '';
+
+      sops.templates.test_template = {
+        content = ''
+          Test value: ${config.sops.placeholder.test_key}
+        '';
+        # This path is inside of `/run/secrets`, which isn't allowed.
+        path = "/run/secrets/foo-bar";
```
Mic92, line 355: If it's inside /run/secrets maybe it should name the template differently rather than doing the symlink? Alternative would be not allow this behavior and make it an eval error.

Mic92:
On the other hand this could also happen with any other invalid symlink as a target.

pkgs/sops-install-secrets/nixos-test.nix
jfly: Ooh, I like the "name it differently" idea. I think that'll actually clean up the implementation a bit, as it currently deals carefully (and verbosely) with the "rendered" subdir.

I'll give this a shot later today.

jfly:
> On the other hand this could also happen with any other invalid symlink as a target.

Sorry, what does "invalid symlink as a target" mean? 

The issue I'm demonstrating here is showing how `/run/secrets` can end up being a directory rather than a symlink.

I suppose there are more subtle scenarios involving symlinks to `/run/secrets` that would not be handled by the fix. I'm not sure if that's what you're referring to.

Issue #665, opened 2024-11-17
Regression: home-manager module fails to evaluate trying to access nonexistant template config


DarkKirb:
Updating to eee831aadbc25f72fa9d214d57bddfd847e026b2 breaks evaluation of my home-manager configurations with the error:

```
[…snip…]
       error: attribute 'templates' missing
       at /nix/store/76d0jm25ajjdlcv33k19jkmvwp1n3dlg-source/modules/home-manager/sops.nix:97:41:
           96|
           97|   manifest = manifestFor "" cfg.secrets cfg.templates;
             |                                         ^
           98|
```

This is because the updated code now requires a sops.templates option which the home-manager module currently does not possess.

PR #668, opened 2024-11-17
define templates for home-manager

Mic92:
fixes https://github.com/Mic92/sops-nix/issues/665

At least now it evals. Haven't tested templates itself yet.


Mic92:
@mergify queue

Issue #681, opened 2024-11-24
Templates with HM "error: attribute 'xdg' missing"


abraverm:
I have tried the new templates support in HM (https://github.com/Mic92/sops-nix/pull/668) but it produces this error when `path` is not specified:
```
       error: attribute 'xdg' missing

       at /nix/store/f7wqa62l66hlnspd2v5mlhl90k7x51vz-source/modules/home-manager/templates.nix:36:30:

           35|                 # Keep this in sync with `RenderedSubdir` in `pkgs/sops-install-secrets/main.go`
           36|                 default = "${config.xdg.configHome}/sops-nix/secrets/rendered/${config.name}";
             |                              ^
           37|               };
```
Also when the path is specified, matching default, the rendered file seems to be encoded, with content like `212476d9b0f3472eaa762...` .

jobs62:
@abraverm actually it is your secret repeated again and again (with one letter in between). maybe you should rotate it, just in case.

Mic92:
Fixed in https://github.com/Mic92/sops-nix/pull/686

PR #686, opened 2024-11-28
try fixing templates on home-manager

jobs62:
Fix for #681, or at least trying.
Should probably be squashed, because i kind of bruteforce the issue

EDIT: i did squashed

### DIFF
--- a/pkgs/sops-install-secrets/main.go
+++ b/pkgs/sops-install-secrets/main.go
@@ -698,7 +698,11 @@ func (app *appContext) validateManifest() error {
 		// The Nix module only defines placeholders for secrets if there are
 		// templates.
 		if len(m.Templates) > 0 {
-			placeholder := m.PlaceholderBySecretName[secret.Name]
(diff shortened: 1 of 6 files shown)


pkgs/sops-install-secrets/main.go
```diff
@@ -698,7 +698,11 @@ func (app *appContext) validateManifest() error {
 		// The Nix module only defines placeholders for secrets if there are
 		// templates.
 		if len(m.Templates) > 0 {
-			placeholder := m.PlaceholderBySecretName[secret.Name]
+			placeholder, present := m.PlaceholderBySecretName[secret.Name]
+			if !present {
+				return fmt.Errorf("placeholder for %s not fount in manifest", secret.Name)
```
Mic92, line 703:
```suggestion
				return fmt.Errorf("placeholder for %s not found in manifest", secret.Name)
```

Mic92:
@mergify queue

PR #698, opened 2024-12-29
Fix link to "more complex .sops.yaml example"

Ten0:
NB: It's unclear to me why it became unnecessary in that repo.
Any insight?


Mic92:
I am using clan now. Which build on top of sops-nix. I can generate secrets declarative with NixOS modules: https://docs.clan.lol/reference/cli/vars/

Also part of our NixCon talk: https://www.youtube.com/watch?v=oAzkHpSWhaI
Secrets are than put in directories to reduce merge conflicts: https://github.com/Mic92/dotfiles/tree/main/vars/per-machine/matchbox2

Mic92:
@mergify queue

Issue #700, opened 2024-12-29
generateKey breaks launchd daemon


Kidswiss:
Hi

If `generateKey` is set, the launchd daemon script will look like this:
```
"/bin/wait4path /nix/store && exec if [[ ! -f '/var/lib/sops-nix/key.txt' ]]; then
  echo generating machine-specific age key...
*snip*
```

https://github.com/Mic92/sops-nix/blob/master/modules/nix-darwin/default.nix#L172
https://github.com/Mic92/sops-nix/blob/master/modules/home-manager/sops.nix#L123

However, the `if` doesn't really work in conjunction with `exec`.

Executing this in bash or zsh will show the issue:
```
exec if [[ ! -f ${escapedAgeKeyFile} ]]; then  
echo test
fi 
zsh: parse error near `then'
```

This breaks the automated mounting of the secrets on startup.

A workaround for now is just setting `generateKey` to false after the first run.

mettavi:
I have exactly the same problem on nix-darwin.

Your solution worked perfectly, thank you so much for posting it.

Mic92:
Is `launchd.daemons.sops-install-secrets` the failing service?

mettavi:
> Is `launchd.daemons.sops-install-secrets` the failing service?

Yes, with `generateKey` set to `true`, `sops-install-secrets` is:

```
"/bin/wait4path /nix/store && exec if [[ ! -f '/var/lib/sops-nix/key.txt' ]]; then
  echo generating machine-specific age key...
*snip*
```
which generates the following error:

```
/bin/sh: -c: line 0: syntax error near unexpected token `then'
/bin/sh: -c: line 0: `/bin/wait4path /nix/store && exec if [[ ! -f /var/lib/sops-nix/key.txt ]]; then'
/bin/sh: -c: line 0: syntax error near unexpected token `then'
/bin/sh: -c: line 0: `/bin/wait4path /nix/store && exec if [[ ! -f /var/lib/sops-nix/key.txt ]]; then'
```

But when `generateKey` is disabled, `sops-install-secrets` is:

```
/bin/sh -c '/bin/wait4path /nix/store && exec 
echo "Setting up secrets..."
*snip*
```
and the error goes away.

Kidswiss:
@Mic92 

The error occurs in the system wide service as well as the user specific one:
* org.nixos.sops-install-secrets-for-users
* org.nixos.sops-install-secrets

Mic92:
Please check the pull request

PR #702, opened 2025-01-02
nix-darwin: fix launchd decrypt scripts

Mic92:
Untested, potentially fixes https://github.com/Mic92/sops-nix/issues/700

please test: @Kidswiss @mettavi

### DIFF
--- a/modules/nix-darwin/default.nix
+++ b/modules/nix-darwin/default.nix
@@ -384,7 +384,7 @@ in
       };
 
       launchd.daemons.sops-install-secrets = {
-        command = installScript;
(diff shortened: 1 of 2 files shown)


mettavi:
With this commit, `org.nixos.sops-install-secrets` (and `sops-install-secrets-for-users`) is now:

```
/bin/sh -c '/bin/wait4path /nix/store && exec sh -c \'
echo "Setting up secrets..."
```

And with the `generateKeys` option enabled and then rebooting, the secrets are successfully mounted at startup.

It fixed it, thank you @Mic92.

Kidswiss:
Works for me too, thanks!

PR #703, opened 2025-01-03
modules/nix-darwin/secrets-for-users: empty set instead of empty list

zowoq:
These were enabled even if there wasn't any user secrets.

Also `SOPS_GPG_EXE` is enabled and gpg, etc is added to the system closure as `sops.gnupg.sshKeyPaths` is set by default, should I just set `sops.gnupg.sshKeyPaths = [ ]` downstream?

### DIFF
--- a/modules/nix-darwin/secrets-for-users/default.nix
+++ b/modules/nix-darwin/secrets-for-users/default.nix
@@ -37,11 +37,11 @@ in
     }
   ];
 
-  system.activationScripts = lib.mkIf (secretsForUsers != [ ]) {
+  system.activationScripts = lib.mkIf (secretsForUsers != { }) {
     postActivation.text = lib.mkAfter installScript;
   };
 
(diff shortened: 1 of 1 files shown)


Mic92:
@mergify queue

Mic92:
> These were enabled even if there wasn't any user secrets.
> 
> Also `SOPS_GPG_EXE` is enabled and gpg, etc is added to the system closure as `sops.gnupg.sshKeyPaths` is set by default, should I just set `sops.gnupg.sshKeyPaths = [ ]` downstream?

You mean because it imports the rsa ssh host key?

zowoq:
Yes, rsa key is imported but isn't needed.

zowoq:
@Mic92 seems CI is stuck?

Mic92:
https://github.com/NixOS/nixpkgs/pull/370959 that is fixed now.

Mic92:
@mergify rebase

Issue #149, opened 2021-12-16
Secrets not showing up in /run/secrets after reboot


DanielFabian:
I just followed the tutorial and tried setting up wifi for my nixos install using nix-sops and it seems to work just fine when I make a `nixos-rebuild switch`. 

However, when I reboot, for some reason, the `/run/secrets` directory isn't created. Doing another `nixos-rebuild switch` (without any changes at all), fixes it again until I reboot.

https://github.com/DanielFabian/.dotfiles/commit/230b4ecf2b017252926f680b7091a94f8634da6f

Any idea what I might be doing wrong? Or is it perhaps a bug that it's not setting up the secrets during the boot.

Mic92:
Can you get boot logs? Is your home that has your key stored on a different partition that is not present yet during boot?

0x4A6F:
I had my secrets located in a different ZFS dataset (!=rootfs) and thus also experienced this behaviour.
The ZFS dataset (mounted in 'Local File Systems' systemd target) was mounted after `stage-2-init: setting up secrets...`.

After moving my sops secrets to my rootfs it does set up secrets during boot.

DanielFabian:
ah, interesting, I'll look at it. And to answer that, yes, totally. In ZFS I have tons of datasets. In particular my user's home directory is a dataset.

Mic92:
Ok. Unfortunately we cannot handle this usecase right now since we need to run before any additional mounts trigged by zfs or by systemd mount units are loaded.

DanielFabian:
so which file exactly do I have to move then? Is it the encrypted sops file, or what?

Mic92:
You need to move your age key: https://github.com/DanielFabian/.dotfiles/blob/230b4ecf2b017252926f680b7091a94f8634da6f/system/configuration.nix#L22

DanielFabian:
thanks, I'll give it a go.

DanielFabian:
thanks, this worked. In the end, I just use the host's SSH key and it's fine.

Mic92:
I think we should mention this in documentation at least.

drestrepom:
I have the same problem, I am trying to add an ssh key in the shell init, but after reboot this not works
```sh
/run/secrets/ssh_key: No such file or directory
```

```nix
  programs.zsh = {
    enable = true;
    autosuggestions.enable = true;
    interactiveShellInit = ''
      export DIRENV_WARN_TIMEOUT=1h
[...]

PR #714, opened 2025-01-17
docs: expand a bit on user secrets + impermanence.

nicdumz:
I spent too much time trying to understand why secrets were not available at login.

The discussion at https://github.com/Mic92/sops-nix/issues/149 was enlightening and provided the last fix (=make sure that whatever `fileSystems."/where/decryption/keys/are".neededForBoot` is set), so documenting it.


Mic92:
Thanks!

PR #715, opened 2025-01-18
Uid/gid for templates

### LINKED ISSUE #514: setting owner and group seems not to work for ids

dkowis:
Inspired by #637 

I'm *almost* done with this, but I need some help. I can't figure out how to get assertions into it. I'm not quite skilled enough with nix to figure that part out on my own. Happy to squash commits when done.

This should resolve #514 by adding uid/gids to templates.

### DIFF
--- a/modules/nix-darwin/templates/default.nix
+++ b/modules/nix-darwin/templates/default.nix
@@ -49,18 +49,32 @@ in
                 '';
               };
               owner = mkOption {
-                type = types.singleLineStr;
-                default = "root";
+                type = with lib.types; nullOr singleLineStr;
+                default = null;
                 description = ''
-                  User of the file.
+                  User of the file. Can only be set if uid is 0;
+                '';
+              };
+              uid = mkOption {
+                type = with lib.types; nullOr int;
+                default = 0;
+                description = ''
+                  UID of the template, only applied with owner is null. the UID will be applied even if the corresponding user doesn't exist.
                 '';
               };
               group = mkOption {
-                type = types.singleLineStr;
-                default = "staff";
+                type = with lib.types; nullOr singleLineStr;
+                default = if config.owner != null then "staff" else null;
                 defaultText = "staff";
                 description = ''
-                  Group of the file. Default on darwin in staff.
+                  Group of the file. Can only be set if gid is 0. Default on darwin to 'staff'
+                '';
+              };
+              gid = mkOption {
+                type = with lib.types; nullOr int;
+                default = 0;
+                description = ''
+                  GID of the template, only applied when group is null. The GID will be applied even if the corresponding group doesn't exist.
                 '';
               };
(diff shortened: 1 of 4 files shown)


modules/sops/templates/default.nix
```diff
@@ -52,18 +52,32 @@ in
...
               owner = mkOption {
-                type = types.singleLineStr;
-                default = "root";
+                type = with lib.types; nullOr singleLineStr;
+                default = null;
+                description = ''
+                  User of the file. Can only be set if uid is 0;
+                '';
+              };
+              uid = mkOption {
```
Mic92, line 61: Can you also add those to `sops-nix/modules/nix-darwin/default.nix`?

dkowis: No problem!

modules/nix-darwin/templates/default.nix
```diff
@@ -49,18 +49,32 @@ in
...
                 description = ''
-                  User of the file.
+                  UID of the template, only applied with owner is null. the UID will be applied even if the corresponding user doesn't exist.
                 '';
               };
               group = mkOption {
-                type = types.singleLineStr;
-                default = "staff";
+                type = with lib.types; nullOr singleLineStr;
+                default = if config.owner != null then "staff" else null;
```
Mic92, line 67: I assume you have not tested this on macOS yet?
I just need to know if I need to give this a try before merging.

dkowis: That is correct, I do not have an OSX I can test it with.

dkowis: Also, just in case you didn't see it, I still haven't figured out how to add assertions, if you could offer a bit of advice there, I'd appreciate it :D

Mic92: Assertions should be fixed.

Mic92:
@dkowis rebased and fixed assertions. Also tested on macOS. Can you have a look again?

dkowis:
Just verified! Works great for me :D

PR #718, opened 2025-01-26
fix import keys hook using unbound variable

### LINKED ISSUE #372: sops-import-keys-hook.bash uses unbound var
On line 3,
```bash
  if [ -n "${sopsCreateGPGHome}" ]; then
```
uses `sopsCreateGPGHome`, which may be unbound (unless the user has set it). Since this file gets sourced into shells, it can break when it’s sourced into a shell with `nounset` (`set -u`) enabled (which I have just run into).

This should either also check that the variable is bound
```bash
  if [ -v sopsCreateGPGHome && -n "${sopsCreateGPGHome}" ]; then
```
or should bind it preemptively
```bash
sopsCreateGPGHome=
…
  if [ -n "${sopsCreateGPGHome}" ]; then
```

0x5a4:
closes #372

this is an issue for me with `nix-direnv`. i've also tried the suggested solution and it actually does work.

### DIFF
--- a/pkgs/sops-import-keys-hook/sops-import-keys-hook.bash
+++ b/pkgs/sops-import-keys-hook/sops-import-keys-hook.bash
@@ -1,6 +1,6 @@
 sopsImportKeysHook() {
   local key dir
-  if [ -n "${sopsCreateGPGHome}" ]; then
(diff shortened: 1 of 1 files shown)


Mic92:
@mergify queue

0x5a4:
@Mic92 i feel like the bot is broken?

PR #722, opened 2025-02-10
Fix symlink ownership change on Darwin systems.

### LINKED ISSUE #701: Symlinks to other directories: "cannot chown symlink/too many levels of symbolic links"
Hi, I am running sops-nix on nix-darwin, with home-manager as a nix-darwin module, using flakes.

Whenever I set the `path =` option on a secret to create a symlink, I am getting an error. For example, I  set the following options on one secret:

```
github_token = {
  owner = "${config.users.users.ta.name}";
  path = "${config.users.users.ta.home}/.config/sops/age/github_token";
  };
```
Then when I run `darwin-rebuild switch` I get the following error on activation:

```
Setting up secrets...
sops-install-secrets: Imported /etc/ssh/ssh_host_ed25519_key as age key
with fingerprint xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
/nix/store/rrlk1kncx6mvzaqb8fz5fnqpfcbpma3p-sops-install-secrets-0.0.1/bin/sops-install-secrets: 
failed to prepare symlinks to secret store: failed to symlink secret '/Users/timotheos/.config/sops/age/github_token': 
cannot chown symlink '/Users/timotheos/.config/sops/age/github_token': 
failed to open /Users/timotheos/.config/sops/age/github_token: too many levels of symbolic links
```
My home-manager sops config then doesn't get activated.

However, inspecting the symlink directory shows that the symlink has indeed been created.

```
 ls -al ~/.config/sops/age
total 4
drwxr-xr-x 5 timotheos staff 160 Jan  1 17:57 .
drwxr-xr-x 3 timotheos staff  96 Dec 30 14:46 ..
lrwxr-xr-x 1 root      staff  25 Jan  1 17:57 github_token -> /run/secrets/github_token
-rw------- 1 timotheos staff 189 Dec 30 14:48 keys.bak
lrwxr-xr-x 1 root      staff  27 Jan  1 17:38 keys.txt -> /run/secrets/encryption_key
```
And the actual file:

```
 sudo ls -al /run/secrets/
total 12
drwxr-x--x 2 root      wheel 170 Jan  1 17:35 .
drwxr-xr-x 4 root      wheel 272 Jan  1 17:57 ..
-r-------- 1 timotheos staff 149 Jan  1 17:35 cachix_auth_token
-rw-r--r-- 1 timotheos staff 189 Jan  1 17:35 encryption_key
-r-------- 1 timotheos staff  40 Jan  1 17:35 github_token
```
(Running `darwin-rebuild switch` a second time brings no change.)

However, if I change the sops-nix config to NOT change the user, then I can use `path =` without error:

```
      github_token = {
        # owner = "${config.users.users.ta.name}";
        path = "${config.users.users.ta.home}/.config/sops/age/github_token";
      };
```
Resulting in:

```
Setting up secrets...
sops-install-secrets: Imported /etc/ssh/ssh_host_ed25519_key as age key
with fingerprint xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
Setting up secrets for users
sops-install-secrets: Imported /etc/ssh/ssh_host_ed25519_key as age key
with fingerprint xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```
So in other words, it looks like the `owner =` and the `path =` options are mutually exclusive. I can only use one of them at a time.

If anyone can help to diagnose the problem here, I would really appreciate it. Happy to provide further information if needed.

```
 nix-info
system: "x86_64-darwin", multi-user?: yes, version: nix-env (Nix) 2.24.10,
channels(root): "", nixpkgs: /nix/store/w5fqkpwgg6q0iw7c5sd9nab8dq94wapk-source
```

ddormer:
Currently trying to set ownership of a symlink on MacOS fails with the error: `too many levels of symbolic links`, which is expected behaviour when trying to open a symlink with the flag `O_NOFOLLOW` on MacOS/Darwin; On Linux systems, the `O_NOFOLLOW` can be combined with the `O_PATH` flag, to access a symlink directly but on Darwin systems the combination and the `O_PATH` flag itself doesn't exist; The flag `O_SYMLINK` can be used instead to reference a symlink directly, instead of following it to the target file.

Once the symlink file descriptor reference is successfully obtained, I found that the `Readlinkat` code, as well as the `Fchownat` is also not Darwin-compatible.

In the case of `Readlinkat`, passing a path to a file as the first argument and an empty string as the second is considered normal practice when using the above `O_NOFOLLOW|O_PATH` Linux method, but isn't available on Darwin either, as an error with the message `not a directory` is raised. Using `Readlink` instead, will correctly provide the symlink target.

In the `Fchownat` case, we see a similar error message "not a directory", and can use `Fchown` and pass the file descriptor that references the symlink itself to the function, providing the same security feature as the Linux implementation, as `Fchown` won't follow the symlink from the given file descriptor.

Fixes #701

### DIFF
--- a/pkgs/sops-install-secrets/darwin.go
+++ b/pkgs/sops-install-secrets/darwin.go
@@ -9,8 +9,6 @@ import (
 	"os"
 	"os/exec"
 	"strings"
-
-	"golang.org/x/sys/unix"
 )
 
 func RuntimeDir() (string, error) {
@@ -21,29 +19,6 @@ func RuntimeDir() (string, error) {
 		return "", fmt.Errorf("cannot get DARWIN_USER_TEMP_DIR: %v", err)
 	}
 	return strings.TrimSuffix(rundir, "/"), nil
-}
-
-func SecureSymlinkChown(symlinkToCheck string, expectedTarget string, owner, group int) error {
-	// not sure what O_PATH is needed for anyways
-	fd, err := unix.Open(symlinkToCheck, unix.O_CLOEXEC|unix.O_SYMLINK|unix.O_NOFOLLOW, 0)
-	if err != nil {
-		return fmt.Errorf("failed to open %s: %w", symlinkToCheck, err)
-	}
-	defer unix.Close(fd)
-
-	buf := make([]byte, len(expectedTarget)+1) // oversize by one to detect trunc
-	n, err := unix.Readlinkat(fd, "", buf)
-	if err != nil {
-		return fmt.Errorf("couldn't readlinkat %s", symlinkToCheck)
-	}
-	if n > len(expectedTarget) || string(buf[:n]) != expectedTarget {
-		return fmt.Errorf("symlink %s does not point to %s", symlinkToCheck, expectedTarget)
-	}
-	err = unix.Fchownat(fd, "", owner, group, unix.AT_SYMLINK_NOFOLLOW)
-	if err != nil {
-		return fmt.Errorf("cannot change owner of '%s' to %d/%d: %w", symlinkToCheck, owner, group, err)
-	}
-	return nil
 }
 
 // Does:
--- a/pkgs/sops-install-secrets/linux.go
+++ b/pkgs/sops-install-secrets/linux.go
@@ -18,38 +18,6 @@ func RuntimeDir() (string, error) {
 	return rundir, nil
 }
 
-func SecureSymlinkChown(symlinkToCheck, expectedTarget string, owner, group int) error {
-	// fd, err := unix.Open(symlinkToCheck, unix.O_CLOEXEC|unix.O_PATH|unix.O_NOFOLLOW, 0)
-	fd, err := unix.Open(symlinkToCheck, unix.O_CLOEXEC|unix.O_PATH|unix.O_NOFOLLOW, 0)
-	if err != nil {
-		return fmt.Errorf("failed to open %s: %w", symlinkToCheck, err)
-	}
-	defer unix.Close(fd)
-
-	buf := make([]byte, len(expectedTarget)+1) // oversize by one to detect trunc
-	n, err := unix.Readlinkat(fd, "", buf)
-	if err != nil {
-		return fmt.Errorf("couldn't readlinkat %s", symlinkToCheck)
-	}
-	if n > len(expectedTarget) || string(buf[:n]) != expectedTarget {
-		return fmt.Errorf("symlink %s does not point to %s", symlinkToCheck, expectedTarget)
-	}
-	stat := unix.Stat_t{}
-	err = unix.Fstat(fd, &stat)
-	if err != nil {
-		return fmt.Errorf("cannot stat '%s': %w", symlinkToCheck, err)
-	}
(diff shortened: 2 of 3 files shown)


pkgs/sops-install-secrets/darwin.go
```diff
@@ -24,22 +24,35 @@ func RuntimeDir() (string, error) {
...
+		unix.O_RDONLY)
 	if err != nil {
 		return fmt.Errorf("failed to open %s: %w", symlinkToCheck, err)
 	}
 	defer unix.Close(fd)
 
+	// Verify that the symlink points to the expected target file.
 	buf := make([]byte, len(expectedTarget)+1) // oversize by one to detect trunc
-	n, err := unix.Readlinkat(fd, "", buf)
+	n, err := unix.Readlink(symlinkToCheck, buf)
```
Mic92, line 43: The whole point of this function is to operate on a fd rather than a path that can change. However changing this function to Readlink would introduce a TOCTOU attack, at which point we could also not check it.

Mic92: However I am currently also wondering if we actually need to make sure that the symlink points to the right target.

Mic92: Maybe instead of `Readlinkat` we can at least make sure with fstat that this is a still symlink? I think the code was originally inspired by systemd-tmpfiles: https://github.com/systemd/systemd/blob/fd2a114061786005f6f088652b100341165ef897/src/tmpfiles/tmpfiles.c#L2459

ddormer: Thanks for the review, it's much appreciated!

I suppose if we didn't check that the fd points to the correct target, then we might have a similar timing problem in between the creation of the symlink here https://github.com/Mic92/sops-nix/blob/master/pkgs/sops-install-secrets/main.go#L192 and the file open in `SecureSymlinkChown`.

Your fstat suggestion sounds reasonable to me, I'll make the changes and test it on my system.

ddormer: After looking over this a bit, `unix.Fstat` doesn't follow the symlink fd, so it doesn't seem like we'd be able to compare the files this way, unless I'm missing something.

I'll look into the subdirectory suggestion in the meanwhile.

ddormer: I've implemented the root-owned subdirectory approach. I've also replaced both of the Linux and Darwin implementations with this single implementation, however if you'd prefer to have the subdirectory approach be Darwin-only and leave the existing Linux implementation alone, I can revert some of the changes.

I think the latest build failed due to a buildbot infrastructure problem?

ddormer:
@Mic92 Sorry for the ping, I don't mind waiting for as long as you'd like for a re-review, I'm just checking if pushing the "request review" last month was enough, or if that doesn't do anything.

Mic92:
Yes. It was enough. I just came back from a longer trip and had to catch up with quite a few projects.

PR #762, opened 2025-04-01
make writing templates more atomic

Mic92:
If the activation script is called twice for some reason, we might end up with a file potentially opened up and written twice.

### DIFF
--- a/pkgs/sops-install-secrets/main.go
+++ b/pkgs/sops-install-secrets/main.go
@@ -198,7 +198,7 @@ func SecureSymlinkChown(targetFile string, path string, owner int, group int) er
 	defer os.RemoveAll(dir)
 
 	// Create symlink to `targetFile` in the temp dir, before chowning it.
-	var tmpSymlink = filepath.Join(dir, filepath.Base(path))
+	tmpSymlink := filepath.Join(dir, filepath.Base(path))
 	if err = os.Symlink(targetFile, tmpSymlink); err != nil {
 		return fmt.Errorf(
 			"cannot create symlink '%s' (pointing to '%s'): %w", path, targetFile, err)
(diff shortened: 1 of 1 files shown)


pkgs/sops-install-secrets/main.go
```diff
@@ -1196,21 +1196,59 @@ func replaceRuntimeDir(path, rundir string) (ret string) {
...
 
-		if err := createParentDirs(targetDir, template.Name, keysGID, userMode); err != nil {
-			return err
+	if err := createParentDirs(targetDir, template.Name, keysGID, userMode); err != nil {
+		return err
+	}
+
+	dir := filepath.Dir(templatePath)
+	tempFile, err := os.CreateTemp(dir, "sops-*")
+	tempfileRemoved := false
```
brianmcgee, line 1208: `tempfileRenamed` might be a better name for this flag.

pkgs/sops-install-secrets/main.go
```diff
@@ -1196,21 +1196,59 @@ func replaceRuntimeDir(path, rundir string) (ret string) {
...
+	defer func() {
+		tempFile.Close() // noop if already closed
+
+		if !tempfileRemoved {
+			if err := os.Remove(tempFile.Name()); err != nil {
+				fmt.Fprintf(os.Stderr, "failed to remove temporary file %s: %s\n", tempFile.Name(),
+					err)
+			}
 		}
+	}()
```
brianmcgee, line 1221: I like to reduce nested ifs and keep the control flow as linear as possible. 

```suggestion
        if tempFileRemoved {
            // nothing to do
            return
        }
		
        if err := os.Remove(tempFile.Name()); err != nil {
        	fmt.Fprintf(os.Stderr, "failed to remove temporary file %s: %s\n", tempFile.Name(),
        		err)
        }		
	}()
```

brianmcgee (changes requested):

PR #680, opened 2024-11-23
feat: add age plugin and fido2 hmac support

brianmcgee:

PR #781, opened 2025-04-24
feat: add age plugin and fido2 hmac support, take 2

NovaViper:
Continuing directly off of #680 .

 I went and updated the changes to work with the most recent commits of sops-nix as well as adding in the wrapper changes introduced into the sops and age packages with https://github.com/NixOS/nixpkgs/pull/395189. I can confirm it does work when you rebuild the NixOS install (and home-manager) but I hadn't been able to figure out how to get sops-nix to install the secrets on boot with the yubikey secrets. I'm using the `age-plugin-yubikey` plugin on my system but all other plugins should work fine.
Additionally, I also went and added the plugin support over in the home-manager module too! I don't have a Mac to test the nix-darwin changes but I'm sure it's simply just adding the same changes into the darwin module aswell.

Some things I've noticed also I want to mention:
- Using keys that *doesn't* require a pin seems to work more consistently (especially for the home-manager NixOS module) than those that do. From what I can tell, it's because the installer program's running in tty which doesn't have keyboard access
- Putting the public keys generated from the plugins in a `keys.txt` file is crucial otherwise sops can't figure out what keys to access. And I've noticed you can put multiple public keys in the same file so that's what I did for all of my age keys from my yubikeys
- Changing the keyFile path without rebuilding the configs first can cause you to get locked out of the secrets (the sops command itself will just refuse to decrypt the secrets). This I found out the hard way 😭


modules/home-manager/sops.nix
```diff
@@ -240,6 +240,16 @@ in
         '';
       };
 
+      plugins = lib.mkOption {
+        type = lib.types.listOf lib.types.package;
+        default = [
+          pkgs.age-plugin-fido2-hmac
```
Mic92, line 246: I think, I don't want to set a default here for the first iteration. Especially this plugin might not work very well since we cannot enter a PIN easily - Maybe this works for activation scripts but with systemd services I am not sure how this is supposed to work.

modules/sops/secrets-for-users/default.nix
```diff
@@ -34,8 +34,9 @@ in
       {
         wantedBy = [ "systemd-sysusers.service" ];
         before = [ "systemd-sysusers.service" ];
-        environment = cfg.environment;
+        environment = cfg.environment // {PATH = lib.mkForce "${cfg.environment.PATH}:${lib.makeSearchPathOutput "bin" "sbin" cfg.age.plugins}";};
```
Mic92, line 37: cc @brianmcgee for review.

modules/home-manager/sops.nix
Mic92: I think this pull request should at least describe how one can setup one age plugin of choice from start to end.

modules/sops/default.nix
```diff
@@ -432,12 +442,15 @@ in
...
       );
 
+      sops.environment.PATH = lib.makeBinPath cfg.age.plugins;
+
       # When using sysusers we no longer are started as an activation script because those are started in initrd while sysusers is started later.
       systemd.services.sops-install-secrets = lib.mkIf (regularSecrets != { } && useSystemdActivation) {
         wantedBy = [ "sysinit.target" ];
         after = [ "systemd-sysusers.service" ];
-        environment = cfg.environment;
+        environment = cfg.environment // {PATH = lib.mkForce "${cfg.environment.PATH}:${lib.makeSearchPathOutput "bin" "sbin" cfg.age.plugins}";};
```
Mic92, line 451: Should this not be in the paths systemd option instead?

modules/home-manager/sops.nix
NovaViper: > I think, I don't want to set a default here for the first iteration. Especially this plugin might not work very well since we cannot enter a PIN easily - Maybe this works for activation scripts but with systemd services I am not sure how this is supposed to work.

Yeah, I can leave that option blank. 👍🏾 

Speaking of systemd.. so I was trying userborn on my laptop to see if the plugins (specifically the yubikey one) works but with the sops-install-secrets-for-users service, I'm seeing warnings about the HOME variable isn't defined and it fails to interact with the key (the key flashes on like it got a request to do something but fails to do actually do the request)
```
● sops-install-secrets-for-users.service
     Loaded: loaded (/etc/systemd/system/sops-install-secrets-for-users.service; enabled; preset: ignored)
     Active: active (exited) since Fri 2025-05-16 10:47:05 CDT; 50min ago
 Invocation: c65e2068806247bb8ed3193542af0afc
    Process: 34406 ExecStart=/nix/store/yh0g1mxx2zcv4snnrgblxdss9mjbxbs5-sops-install-secrets-0.0.1/bin/sops-install-secrets -ignore-passwd /nix/store/b9swyn2kxpf98cdycg6yhshjm62131vw-manifest-for-users.json (code=exited, status=0/SUCCESS)
   Main PID: 34406 (code=exited, status=0/SUCCESS)
         IP: 0B in, 0B out
         IO: 0B read, 0B written
   Mem peak: 8.9M
        CPU: 60ms

May 16 10:47:05 yoganova systemd[1]: Starting sops-install-secrets-for-users.service...
May 16 10:47:05 yoganova sops-install-secrets[34406]: sops-install-secrets: Imported /etc/ssh/ssh_host_ed25519_key as age key with fingerprint age1XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
May 16 10:47:05 yoganova sops-install-secrets[34406]: [AGE]         time="2025-05-16T10:47:05-05:00" level=warning msg="could not determine the user home directory: $HOME is not defined"
May 16 10:47:05 yoganova sops-install-secrets[34406]: [AGE]         time="2025-05-16T10:47:05-05:00" level=warning msg="could not determine the user home directory: $HOME is not defined"
May 16 10:47:05 yoganova sops-install-secrets[34406]: [AGE]         time="2025-05-16T10:47:05-05:00" level=warning msg="could not determine the user home directory: $HOME is not defined"
May 16 10:47:05 yoganova sops-install-secrets[34406]: [AGE]         time="2025-05-16T10:47:05-05:00" level=warning msg="could not determine the user home directory: $HOME is not defined"
May 16 10:47:05 yoganova systemd[1]: Finished sops-install-secrets-for-users.service.
```

Even stranger is the home-manager sops-nix service. It just completely skips the key, similarly flashing on like how I described with the install secrets service. But as soon as I turn off userborn, the home-manager service works fine for the home-manager service!

🤔 What's really odd is I've been able to get the plugin working over in agenix with home-manger but not with the user passwords (similarly had issues with the plugin not being executed during the boot process, always had to include an ssh key as a fallback key to decrypt the user passwords)

NovaViper: > I think this pull request should at least describe how one can setup one age plugin of choice from start to end.

Will do! I intended to do one earlier but I forgot to do so 😭

NovaViper: > Even stranger is the home-manager sops-nix service. It just completely skips the key, similarly flashing on like how I described with the install secrets service. But as soon as I turn off userborn, the home-manager service works fine for the home-manager service!

Ah I realized. The age key I was using on my yubikey had a PIN policy on it, switched that one out for a pinless one and now the home-manager service doesn't break anymore (with either userborn and standard! 🥳 ); but I still can't figure out that `$HOME is not defined` warning with the NixOS service 🤔

NovaViper:
I also withdrew the darwin changes in case those caused breakages over on that side (I can't test that build at all since I don't have a Mac)

NovaViper:
And the write up is done!

NovaViper:
I completely missed the `withEnvironment` function that's used for the activation script in my last rebase, which caused the activation script not to use the plugins! 👀 
Adding the plugins to the PATH environment variable there makes the plugins work again.

modules/sops/default.nix
```diff
@@ -438,6 +447,7 @@ in
         after = [ "systemd-sysusers.service" ];
         environment = cfg.environment;
         unitConfig.DefaultDependencies = "no";
+        path = cfg.age.plugins;
```
sean-xyz, line 450:
```suggestion
        path = lib.mkForce cfg.age.plugins;
```

this was causing errors in `nix flake check`, with
> [environment.](error: The option `systemd.services.sops-install-secrets.environment.PATH' has conflicting definition values)

modules/sops/secrets-for-users/default.nix
```diff
@@ -36,6 +37,7 @@ in
         before = [ "systemd-sysusers.service" ];
         environment = cfg.environment;
         unitConfig.DefaultDependencies = "no";
+        path = cfg.age.plugins;
```
sean-xyz, line 40:
```suggestion
        path = lib.mkForce cfg.age.plugins;
```

this was causing errors in `nix flake check`, with
> [environment.](error: The option `systemd.services.sops-install-secrets-for-users.environment.PATH' has conflicting definition values)

sean-xyz (commented):
editing to delete - i was getting errors that i am not reproducing now

NovaViper:
> after having success with age-plugin-tpm, i'm now trying age-plugin-fido2-hmac and having less success. in particular, for nixos secrets loaded at boot, with or without `neededForUsers`. i see in your comments you're primarily using age-plugin-yubikey, have you had success with other plugins as well @NovaViper ?

Hey sorry for the **extremely** delayed reply! The notification got buried in my emails 😭 I hadn't tested any of the other age plugins, just only the yubikey one.

> when you used piv, were you able to test it with user presence?

And yeah, all of my age keys on my yubikeys are set to use a touch confirmation. I found that having a PIN confirmation seemed to cause issues during boot but the touch based one works the most reliably.

NovaViper:
👀 Also looking I'm a good bit behind with the upstream repo, will rebase it shortly

NovaViper:
> age-plugin-fido2-hmac always requires user presence (e.g. touching the key), while piv (age-plugin-yubikey) doesn't. when you used piv, were you able to test it with user presence? i'm not sure if my problem is the presence check (i don't see my key flashing to prompt me) or fido2 prerequisites not loading in time
> 
> btw, in my view age-plugin-tpm has a nearly identical use case to the [ssh host key route](https://github.com/Mic92/sops-nix/blob/master/README.md?plain=1#L243), but with the added security benefit of TPM for leaked material. the piv or ctap2 plugins may also have some use in a live-image, though i'm not sure how it can be done without exposing identity files.

Now the identity files.. I particularly didn't use the fido2 plugin because of it exposes the identity files directly, vs the yubikey one only showing the public key. I have the identity files located under `~/.config/sops/age/keys.txt`, which allows the base sops detect the keys and perform the requests.

> the piv or ctap2 plugins may also have some use in a live-image, though i'm not sure how it can be done without exposing identity files.

🤔 This I'm not too sure if you can without in some way placing down the identity files somewhere. I notice without it, sops literally has no idea where the keys are, it always looks at `~/.config/sops/age/keys.txt`.

NovaViper:
@rsdy Yeah you do need pcscd enabled! Also did you make sure to copy the identity files over into `~/.config/sops/age/keys.txt`? Do you have any logs for it while you were trying to get the setup going?




> It seems like the overlay may not actually be needed, since you set `path` for the service anyway

The overlay is for the actual command that you access in the terminal when you are setting up the secrets. The service path is something different altogether. The path being set there is only applied to the environment the service will run in, not the entire system (basically anywhere that you'll run sops when you go to create the secrets as a user).

Ramblurr:
I'd also like to give this a test, but is there some consequence to setting enabling and then potentially disabling userborn (if the test doesn't work out)? I don't want to lockmyself out the system heh.

NovaViper:
No, it will use the activation script as normal (that's what I use on my system)

Sep 29, 2025 11:07:05 AM Casey Link ***@***.***>:

PR #826, opened 2025-08-25
sops-install-secrets: use noswap mount option with tmpfs

Sporif:
I think we can assume the `noswap` tmpfs mount option is available, as Linux 6.4 was released 2 years ago. 

I had to use another package to check if the secrets directory is a mountpoint, since only checking if the filesystem type is tmpfs doesn't work when the parent directory `/run` is usually tmpfs. 

## Summary by CodeRabbit

- New Features
  - More reliable secret-mount behavior with a safer, more resilient mounting flow.
- Bug Fixes
  - Improved mount detection, fallback handling, and clearer error reporting for increased stability.
- Documentation
  - Conditioned tmpfs warning for older Linux kernels to reduce unnecessary alerts.
- Dependencies
  - Added/updated mount-related library to improve mount detection.
- Chores
  - Updated default build metadata (vendor hash).

### DIFF
--- a/modules/sops/default.nix
+++ b/modules/sops/default.nix
@@ -306,7 +306,7 @@ in
         Use tmpfs in place of ramfs for secrets storage.
 
         *WARNING*
-        Enabling this option has the potential to write secrets to disk unencrypted if the tmpfs volume is written to swap. Do not use unless absolutely necessary.
+        On Linux kernels earlier than 6.4, enabling this option has the potential to write secrets to disk unencrypted if the tmpfs volume is written to swap. Do not use unless absolutely necessary.
 
         When using a swap file or device, consider enabling swap encryption by setting the `randomEncryption.enable` option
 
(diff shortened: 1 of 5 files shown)


modules/sops/default.nix
```diff
@@ -305,11 +305,6 @@ in
       description = ''
         Use tmpfs in place of ramfs for secrets storage.
 
-        *WARNING*
-        Enabling this option has the potential to write secrets to disk unencrypted if the tmpfs volume is written to swap. Do not use unless absolutely necessary.
-
-        When using a swap file or device, consider enabling swap encryption by setting the `randomEncryption.enable` option
-
         ```
```
SuperSandro2000, line 308: Either we remove the nix example code, too or keep the explanation.

Mic92 (approved):

sebastianrasor:
> > > I think we can assume the `noswap` tmpfs mount option is available, as Linux 6.4 was released 2 years ago.
> > 
> > 
> > A fallback would be still cool though because we still have people stuck on older LTS kernel i.e. special vendor kernels.
> 
> I added a fallback. I also set the `noexec` flag, as I don't see when you would need executable secrets.

It's not that you might need executable secrets, it's that you might need executable templates. Can we revisit this decision @Sporif @Mic92 ?

sebastianrasor:
I went ahead and opened https://github.com/Mic92/sops-nix/issues/864 for standalone discussion on this topic

PR #829, opened 2025-09-06
fix changing between a secret file and a directory with the same name

### LINKED ISSUE #523: Activation fails with "not a directory" if a file becomes a directory
Suppose you had a secret deployed like this:

```nix
{ sops.secrets.foo = {}; }
```

Now, when deploying a new change with the following secret:

```nix
{ sops.secrets."foo/bar" = {}; }
```

The activation fails like this:

```
sops-install-secrets: Imported /etc/ssh/ssh_host_ed25519_key as age key with fingerprint ...
/nix/store/1dilq2rayqd2alchs0bmhssmwni8mbxv-sops-install-secrets-0.0.1/bin/sops-install-secrets: Cannot request units to restart: open /run/secrets/foo/bar: not a directory
```

hgl:
This fixes #523, which https://github.com/Mic92/sops-nix/pull/707 actually doesn't address.

This also fixes changing `{ sops.secrets."foo/bar" = {}; }` back to `{ sops.secrets.foo = {}; }`.

## Summary by CodeRabbit

- Bug Fixes
  - Transitions where a secret or rendered template changes between a file and a directory are now detected and treated as new items: the system will enqueue automatic reloads/restarts and skip further processing for the affected item to prevent errors and stale state.

- New Features
  - None.

### DIFF
--- a/pkgs/sops-install-secrets/main.go
+++ b/pkgs/sops-install-secrets/main.go
@@ -966,8 +966,8 @@ func handleModifications(isDry bool, logcfg loggingConfig, symlinkPath string, s
 		// Read the old file
 		oldData, err := os.ReadFile(oldPath)
 		if err != nil {
-			if os.IsNotExist(err) {
-				// File did not exist before
+			// File did not exist before or the path changed from a file to a directory or vice versa
+			if os.IsNotExist(err) || errors.Is(err, syscall.ENOTDIR) || errors.Is(err, syscall.EISDIR) {
 				restart = append(restart, secret.RestartUnits...)
 				reload = append(reload, secret.ReloadUnits...)
 				newSecrets[secret.Name] = true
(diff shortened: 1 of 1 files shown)


pkgs/sops-install-secrets/main.go
```diff
@@ -961,8 +961,10 @@ func handleModifications(isDry bool, logcfg loggingConfig, symlinkPath string, s
 		// Read the old file
 		oldData, err := os.ReadFile(oldPath)
 		if err != nil {
-			if os.IsNotExist(err) {
+			if os.IsNotExist(err) || errors.Is(err, syscall.ENOTDIR) || errors.Is(err, syscall.EISDIR) {
 				// File did not exist before
+				// Or it changed from a to a/b (ENOTDIR)
+				// Or it changed from a/b to a (EISDIR)
```
SuperSandro2000, line 967:
```suggestion
			// File did not exist before or the path changed from a file to a directory or vice versa
			if os.IsNotExist(err) || errors.Is(err, syscall.ENOTDIR) || errors.Is(err, syscall.EISDIR) {
```

pkgs/sops-install-secrets/main.go
```diff
@@ -992,8 +994,10 @@ func handleModifications(isDry bool, logcfg loggingConfig, symlinkPath string, s
...
```
SuperSandro2000, line 1000:
```suggestion
			// File did not exist before or the path changed from a file to a directory or vice versa
			if os.IsNotExist(err) || errors.Is(err, syscall.ENOTDIR) || errors.Is(err, syscall.EISDIR) {
```

PR #858, opened 2025-11-15
modules/sops: allow forcing systemd-based activation (attempt #2)

### LINKED ISSUE #856: "error: attribute 'systemd' missing" in commit b80c966e70fa0615352c9596315678df1de75801
With nixpkgs-unstable and today's version of sops-nix I get this error from deploy-rs:

```
       … while checking flake output 'checks'
         at /nix/store/d55y2ylzrn67kqb8rv328qkp72iwcvck-source/flake.nix:64:7:
           63|     {
           64|       checks = builtins.mapAttrs (_system: deployLib: deployLib.deployChecks self.deploy) deploy-rs.lib;
             |       ^
           65|

       … while checking the derivation 'checks.x86_64-linux.deploy-schema'

       … while calling the 'derivationStrict' builtin
         at <nix/derivation-internal.nix>:37:12:
           36|
           37|   strict = derivationStrict drvAttrs;
             |            ^
           38|

       … while evaluating the option `system.build.toplevel':

       … while evaluating definitions from `/nix/store/cjjpqrp9xy44j5fjh3mygyxy3i7mslj2-source/nixos/modules/system/activation/top-level.nix':

       … while evaluating the option `warnings':

       … while evaluating definitions from `/nix/store/cjjpqrp9xy44j5fjh3mygyxy3i7mslj2-source/nixos/modules/system/boot/systemd/tmpfiles.nix':

       … while evaluating the option `boot.initrd.systemd.storePaths':

       … while evaluating definitions from `/nix/store/cjjpqrp9xy44j5fjh3mygyxy3i7mslj2-source/nixos/modules/config/console.nix':

       … while evaluating the option `console.keyMap':

       … while evaluating the option `environment.sessionVariables':

       … while evaluating definitions from `/nix/store/cjjpqrp9xy44j5fjh3mygyxy3i7mslj2-source/nixos/modules/services/desktop-managers/gnome.nix':

       … while evaluating the default value of option `sops.useSystemdActivation`

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: attribute 'systemd' missing
       at /nix/store/2d0a7pxy31p21fzs535hgvn49zbi6l8k-source/modules/sops/default.nix:321:40:
          320|       default =
          321|         (options.systemd ? sysusers && config.systemd.sysusers.enable)
             |                                        ^
          322|         || (options.services ? userborn && config.services.userborn.enable);
       Did you mean system?
```

lschuermann:
By allowing users to optionally force using systemd unit-based activation, they can inject dependencies on other services or mountpoints (for instance, when the age key is not stored on the root file system).

This is effectively the same PR as #855, but with a `defaultText` option added. I've managed to reproduce the original reported issue (using `documentation.nixos.includeAllModules = true;`), and with this PR `sops-nix` evaluates even with this option set.

Closes #856

### DIFF
--- a/modules/sops/default.nix
+++ b/modules/sops/default.nix
@@ -29,10 +29,6 @@ let
   # Currently, all templates are "regular" (there's no support for `neededForUsers` for templates.)
   regularTemplates = cfg.templates;
 
-  useSystemdActivation =
-    (options.systemd ? sysusers && config.systemd.sysusers.enable)
-    || (options.services ? userborn && config.services.userborn.enable);
-
   withEnvironment = import ./with-environment.nix {
     # sops >=3.10.0 now unconditionally searches 
     # for an SSH key in $HOME/.ssh/, introduced in #1692 [0]. Since in the
@@ -319,6 +315,23 @@ in
       '';
     };
 
+    useSystemdActivation = lib.mkOption {
+      type = lib.types.bool;
+      default =
+        (options.systemd ? sysusers && config.systemd.sysusers.enable)
+        || (options.services ? userborn && config.services.userborn.enable);
+      defaultText = lib.literalExpression (
+        "(options.systemd ? sysusers && config.systemd.sysusers.enable) "
+        + "|| (options.services ? userborn && config.services.userborn.enable)"
+      );
+      description = ''
+        Use a systemd unit to install secrets, instead of deploying them using an activation script.
+
+        This option is automatically enabled when systemd-sysusers or userborn are used to manage users and groups.
+        It can also be useful to specify additional dependencies to be satisfied before secrets are installed, such as required mountpoints for SOPS key files.
+      '';
+    };
+
     age = {
       keyFile = lib.mkOption {
         type = lib.types.nullOr pathNotInStore;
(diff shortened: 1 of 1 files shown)


lschuermann:
@Mic92 Sorry for all the noise. @sedlund confirmed that this patch doesn't break evaluation any more, so I think this should be good to go.

I tried extending one of the existing NixOS tests with the respective `documentation.nixos` settings that caused this breakage, but have not been able to reproduce this issue in there---I guess something causes the docs to not be built in the test derivations.

SuperSandro2000 (approved):

rebmit:
Should we make `useSystemdActivation` enabled by default, given that https://github.com/NixOS/nixpkgs/pull/475306 proposes deferring `update-users-groups` to run as a stage-2 systemd service?

Mic92:
That would need a check to make sure its enabled in the nixos version used.

musjj:
Unfortunately, this option does not affect `sops-install-secrets-for-users` for some reason. Is the omission intentional?

lschuermann:
> Unfortunately, this option does not affect `sops-install-secrets-for-users` for some reason. Is the omission intentional?

I didn't intend to omit it, this PR was just to scratch a personal itch. I don't see a good reason for why this shouldn't affect that other service, but I'm just a user of sops-nix and don't know much about its internals.

PR #876, opened 2025-12-16
Added `package` option for `gnupg`.

FabrizioRomanoGenovese:
Added a `package` option for `gnupg`. This is useful, for instance, to allow usage with bleeding-edge GPG, which supports post-quantum encryption algos such as Kyber, defusing the "store now, decrypt later" threat model.

### DIFF
--- a/modules/nix-darwin/default.nix
+++ b/modules/nix-darwin/default.nix
@@ -384,7 +384,7 @@ in
 
     {
       sops.environment.SOPS_GPG_EXEC = lib.mkIf (cfg.gnupg.home != null || cfg.gnupg.sshKeyPaths != [ ]) (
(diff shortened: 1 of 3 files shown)


modules/nix-darwin/default.nix
```diff
@@ -384,7 +384,7 @@ in
 
     {
       sops.environment.SOPS_GPG_EXEC = lib.mkIf (cfg.gnupg.home != null || cfg.gnupg.sshKeyPaths != [ ]) (
-        lib.mkDefault "${pkgs.gnupg}/bin/gpg"
+        lib.mkDefault "${cfg.gnupg.package}/bin/gpg"
```
SuperSandro2000: The package option is missing a bit up.

Issue #881, opened 2025-12-29
sops-install-secrets: `SymlinkPath` is required for `chown` before it's created


nazarewk:
# background

I am extending `sops-nix` setup to `nix-darwin` and I am unable to create secrets where the path is at `/run/configs` due to this error (I copied paths from `launchd` service and calling them directly):

```shell
root@anji:~/ > /nix/store/0v4lp6wlhp3afb67yi6yl5rnskfpqqam-sops-install-secrets-0.0.1/bin/sops-install-secrets /nix/store/aw9xmar1rfwnamsb8s0949a1bsdb2kh6-manifest.json
/nix/store/0v4lp6wlhp3afb67yi6yl5rnskfpqqam-sops-install-secrets-0.0.1/bin/sops-install-secrets: failed to prepare symlinks to secret store: failed to symlink secret '/run/configs/anonymization/work-00-company-name/pattern': cannot chown symlink '/run/configs/anonymization/work-00-company-name/pattern': failed to open /run/configs/anonymization/work-00-company-name/pattern: too many levels of symbolic links
```

with the configuration looking as follows (stripped down repetitive parts because it fails on the first secret):

<details>
<summary>manifest.json</summary>

```shell
kdn@brys ~/d/g/n/nix-configs> TERM=xterm ssh anji cat /nix/store/aw9xmar1rfwnamsb8s0949a1bsdb2kh6-manifest.json | jq '.secrets |= [.[0]] | .templates |= .[0] | del(.placeholderBySecretName)'
{
  "ageKeyFile": "/var/lib/sops-nix/key.txt",
  "ageSshKeyPaths": [],
  "gnupgHome": null,
  "keepGenerations": 1,
  "logging": {
    "keyImport": true,
    "secretChanges": true
  },
  "secrets": [
    {
      "format": "yaml",
      "gid": 0,
      "group": "staff",
      "key": "anonymization/work-00-company-name/pattern",
      "mode": "0444",
      "name": "anonymization/work-00-company-name/pattern",
      "neededForUsers": false,
      "owner": "root",
      "path": "/run/configs/anonymization/work-00-company-name/pattern",
      "sopsFile": "/nix/store/8yf3li79np5lpgm07dc162a44za5qi3z-source/default.unattended.sops.yaml",
      "sopsFileHash": "ddb1d98c8486fef51933abe7cbb6d700c93d80741b9e9c59c12544e8292b9a00",
      "uid": 0
    }
  ],
[...]

PR #882, opened 2025-12-29
sops-install-secrets: create /run/secrets link before chowning it

nazarewk:
fixes https://github.com/Mic92/sops-nix/issues/881, see the issue for the background


SuperSandro2000:
Can we easily add a test for this?

PR #911, opened 2026-02-18
fix(darwin): guard launchctl bootstrap on plist existence

### LINKED ISSUE #910: Launchctl bootstrap failed on darwin with I/O error
When setting up sops-nix for the first time on nixos-darwin. I get the following log output:

```
Starting Home Manager activation
Activating checkAppManagementPermission
Activating checkFilesChanged
Activating checkLinkTargets
Activating writeBoundary
Activating installPackages
replacing old 'home-manager-path'
installing 'home-manager-path'
Activating copyApps
setting up ~/Applications/Home Manager Apps...
Activating createGpgHomedir
Activating linkGeneration
Cleaning up orphan links from /Users/jackschumann
Creating home file links in /Users/jackschumann
Activating generateAgeKey
Activating sops-nix
Boot-out failed: 3: No such process
Bootstrap failed: 5: Input/output error
Try re-running the command as root for richer errors.
```

Im able to work around this by setting the activation phase to come later

```nix
 home.activation.sops-nix = lib.mkForce (lib.hm.dag.entryAfter [ "linkGeneration" ] "");
```

Once this is worked around
a `~/Library/LaunchAgents/org.nix-community.home.sops-nix.plist` exists and the work-around is not needed, making this a little hard to repro.

To repro rm that file and run a home-manager switch.

In case its relevant:
```
❯ nix --version
nix (Nix) 2.31.2+1
```

jackschu:
Closes #910

## Problem

During home-manager activation on darwin, the `launchctl bootstrap` call can fail with `Input/output error` if the LaunchAgent plist hasn't been written yet. Home-manager does create the plist during `setupLaunchAgents` which is just constrained to entryAfter `writeBoundary`, but afaik the activation phase in `sops-nix` doesn't impose any ordering.

Once the plist exists from a prior activation, the issue doesn't recur — making it hard to repro outside of first-time setup (or after removing the plist).

(i did repro this once by factory-resetting my macbook, so it I have seen this occur multiple times outside of manually removing the plist)
## Summary

I guard the `launchctl bootstrap` call with a file existence check on the plist. The `bootout` already swallows errors via `&& true` and its less clear that that line should be targeting any particuylar path.

This means:
- First activation (no plist yet): bootout is a no-op, bootstrap is skipped, secrets are still loaded by home-manager itself which calls bo
- Subsequent activations (plist exists): behavior is unchanged, the agent is bounced as before

## Test plan

- `rm ~/Library/LaunchAgents/org.nix-community.home.sops-nix.plist && home-manager switch` no longer errors
- Normal `home-manager switch` with existing plist still reloads secrets

Note: I did use AI to understand these two repos. I came up with the fix and did test this. I'm trying to do this well, but apologies if my understanding is off anywhere.


SuperSandro2000 (approved):

jerome-benoit (commented):
Minor: `&& true` doesn't suppress the exit code — it only prevents `set -e` from aborting. If `bootout` fails, the compound expression still exits non-zero.

```suggestion
            /bin/launchctl bootout ${domain-target}/org.nix-community.home.sops-nix || true
```

`|| true` guarantees exit 0 regardless of whether bootout succeeds or fails.

PR #928, opened 2026-03-21
fix(systemd): require mounts for encryption keys.

nicdumz:
This helps address issues in https://github.com/nix-community/impermanence/issues/294 and in general also works for https://github.com/nix-community/preservation type of workflows which also rely on systemd mounts.

Without this, it could be that files required for `sops-install-secrets` units were not present or mounted.

### DIFF
--- a/modules/sops/default.nix
+++ b/modules/sops/default.nix
@@ -484,6 +484,12 @@ in
               ExecStart = [ "${cfg.package}/bin/sops-install-secrets ${manifest}" ];
               RemainAfterExit = true;
             };
+            unitConfig.RequiresMountsFor = lib.concatLists [
+              (lib.lists.optional (cfg.gnupg.home != null) cfg.gnupg.home)
+              cfg.gnupg.sshKeyPaths
+              (lib.lists.optional (cfg.age.keyFile != null) cfg.age.keyFile)
+              cfg.age.sshKeyPaths
+            ];
           };
 
       system.activationScripts = {
--- a/modules/sops/default.nix
+++ b/modules/sops/default.nix
@@ -430,56 +430,61 @@ in
   ];
   config = lib.mkMerge [
     (lib.mkIf (cfg.secrets != { }) {
-      assertions =
-        [
-          {
-            assertion =
-              cfg.gnupg.home != null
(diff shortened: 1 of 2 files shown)


nicdumz:
@Mic92 gentle poke, what would you think of this? On my end this was the "unlock" to make sops-nix collaborate with systemd mounts and upstreaming would feel like the right thing to me.

SuperSandro2000 (approved):

nicdumz:
@SuperSandro2000 thanks for the review. Are you able to merge it or do you need @Mic92 for this? Thanks :)

musjj:
Hey, I was about to migrate to this setup, but I realized that `sops-install-secrets-for-users` is gated behind `sysusers`/`userborn`:

https://github.com/Mic92/sops-nix/blob/fbedbb8cb1eef3e42c1d25bd59138e2bd0503364/modules/sops/secrets-for-users/default.nix#L28-L34

You can override this behavior for `sops-install-secrets` using the `useSystemdActivation`:

https://github.com/Mic92/sops-nix/blob/fbedbb8cb1eef3e42c1d25bd59138e2bd0503364/modules/sops/default.nix#L319

But this option doesn't affect `sops-install-secrets-for-users` at all. Is this intentional or an oversight?

nicdumz:
@musjj `useSystemdActivation` is gated by userborn by default, and `sops-install-secrets-for-users` is activated if `useSystemdActivation` is true. The unit has RequiresMountsFor (always).

I don't see an inconsistency, could you expand?

musjj:
`sops-install-secrets-for-users` is not enabled if `useSystemdActivation` is `true` but `userborn`/`sysusers` is not. Meanwhile `sops-install-secrets` will respect the option even if `userborn`/`sysusers` is not enabled.

nicdumz:
useSystemdActivation defaults to userborn or sysusers being enabled so this feels like the same to me :)

musjj:
Yeah but you can force it to be enabled for `sops-install-secrets` even if you're not using `userborn` or `sysusers`, which you can't for `sops-install-secrets-for-users`. But it's okay I've migrated to `userborn` just now :)

PR #936, opened 2026-04-03
sops-install-secrets: call systemctl directly when run as a systemd unit

r-vdp:
When `useSystemdActivation` is enabled, `sops-install-secrets.service`
still writes to `/run/nixos/activation-{restart,reload}-list`, but at
that point, those files will not be looked at anymore, and so instead
they cause services to be restarted on the *next* activation.

NixOS 26.05 also deprecates the activation-*-list mechanism, printing a
warning whenever the files exist, with removal planned for 26.11.

This change detects systemd invocation via `INVOCATION_ID` and calls
`systemctl` directly (`try-restart` / `try-reload-or-restart`, with
`--no-block` to avoid deadlocking the sysinit transaction). The legacy
activation-script path keeps writing the list files for backward
compatibility.

Closes #934

### DIFF
--- a/modules/sops/secrets-for-users/default.nix
+++ b/modules/sops/secrets-for-users/default.nix
@@ -35,7 +35,9 @@ in
       {
         wantedBy = [ "systemd-sysusers.service" ];
         before = [ "systemd-sysusers.service" ];
-        environment = cfg.environment;
+        environment = cfg.environment // {
+          SOPS_RESTART_UNITS_VIA_SYSTEMCTL = "1";
+        };
         unitConfig.DefaultDependencies = "no";
         path = cfg.age.plugins;
 
--- a/pkgs/sops-install-secrets/main.go
+++ b/pkgs/sops-install-secrets/main.go
@@ -8,6 +8,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"os/exec"
 	"os/user"
 	"path"
 	"path/filepath"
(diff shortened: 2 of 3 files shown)


modules/sops/secrets-for-users/default.nix
```diff
@@ -37,7 +37,7 @@ in
         before = [ "systemd-sysusers.service" ];
         environment = cfg.environment;
         unitConfig.DefaultDependencies = "no";
-        path = cfg.age.plugins;
+        path = cfg.age.plugins ++ [ config.systemd.package ];
```
SuperSandro2000: Is this not contained in the default path of a systemd unit already?

pkgs/sops-install-secrets/main.go
```diff
@@ -1020,38 +1021,90 @@ func handleModifications(isDry bool, logcfg loggingConfig, symlinkPath string, s
...
+			// --no-block: we are ordered before sysinit-reactivation.target
+			// with DefaultDependencies=no. Blocking on a normal service
+			// (which has After=sysinit.target) would deadlock the
+			// transaction.
+			args := append([]string{"--no-block", verb}, units...)
+			cmd := exec.Command("systemctl", args...)
+			cmd.Stdout = os.Stderr
+			cmd.Stderr = os.Stderr
+			if err := cmd.Run(); err != nil {
+				return fmt.Errorf("systemctl %s %s: %w", verb, strings.Join(units, " "), err)
```
SuperSandro2000:
```suggestion
				return fmt.Errorf("Running "systemctl %s %s" failed: %w", verb, strings.Join(units, " "), err)
```
maybe?

modules/sops/secrets-for-users/default.nix
r-vdp: I didn't think so, but I'd need to check.

pkgs/sops-install-secrets/main.go
r-vdp: reworded

modules/sops/secrets-for-users/default.nix
r-vdp: Yeah you were right, removed now.

PR #942, opened 2026-04-08
Add args option for age key generation

LucasWagler:
Add a flexible `sops.age.extraGenerateKeyArgs` option to the base/NixOS, home-manager, and darwin modules.

This enables:
- **Post-quantum key generation**: `age-keygen` currently offers option `-pq` for this. This is the primary driver behind the PR.
- **Future flexibility**: If `age-keygen` adds more options in the future, `sops-nix` users can use this escape hatch without waiting for new development.

## :speech_balloon: Seeking Feedback: Option Naming
Here are some options I'm considering:
- Rename `generateKey` to `keyGeneration.enable` & add `keyGeneration.extraArgs`
    - :+1: Explicitly groups the two options together
    - :-1: `keyFile` option is left out, despite also being relevant. `keyFile` impacts key generation, but also has a purpose outside of generation.
- `extraGenerateKeyArgs`
    - :+1: Non-invasive. No end user refactoring to get rid of rename warning.
    - :+1: All three options impacting key generation stay at the same nesting level
    - :-1: Implicit grouping, not explicit
    - :-1: Long/wordy

I like the first option better, but the second option is what is currently implemented. Let me know your thoughts!

### DIFF
--- a/modules/home-manager/sops.nix
+++ b/modules/home-manager/sops.nix
@@ -121,7 +121,7 @@ let
     pkgs.writeShellScript "sops-nix-user" (
       lib.optionalString cfg.age.generateKey ''
         if [[ ! -f ${escapedAgeKeyFile} ]]; then
-          echo generating machine-specific age key...
+          echo generating user-specific age key...
           ${pkgs.coreutils}/bin/mkdir -p $(${pkgs.coreutils}/bin/dirname ${escapedAgeKeyFile})
           # age-keygen sets 0600 by default, no need to chmod.
           ${pkgs.age}/bin/age-keygen -o ${escapedAgeKeyFile} ${lib.escapeShellArgs cfg.age.extraGenerateKeyArgs}
--- a/modules/home-manager/sops.nix
+++ b/modules/home-manager/sops.nix
@@ -124,7 +124,7 @@ let
           echo generating machine-specific age key...
           ${pkgs.coreutils}/bin/mkdir -p $(${pkgs.coreutils}/bin/dirname ${escapedAgeKeyFile})
           # age-keygen sets 0600 by default, no need to chmod.
-          ${pkgs.age}/bin/age-keygen -o ${escapedAgeKeyFile}
+          ${pkgs.age}/bin/age-keygen -o ${escapedAgeKeyFile} ${lib.escapeShellArgs cfg.age.extraGenerateKeyArgs}
         fi
       ''
       + ''
@@ -267,6 +267,15 @@ in
         '';
       };
 
+      extraGenerateKeyArgs = lib.mkOption {
+        type = lib.types.listOf lib.types.str;
+        default = [ ];
+        example = [ "-pq" ];
+        description = ''
+          List of arguments to use when generating the age key.
+        '';
+      };
+
       sshKeyPaths = lib.mkOption {
         type = lib.types.listOf lib.types.path;
         default = [ ];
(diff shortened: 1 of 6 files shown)


modules/home-manager/sops.nix
```diff
@@ -121,10 +121,10 @@ let
     pkgs.writeShellScript "sops-nix-user" (
       lib.optionalString cfg.age.generateKey ''
         if [[ ! -f ${escapedAgeKeyFile} ]]; then
-          echo generating machine-specific age key...
+          echo generating user-specific age key...
           ${pkgs.coreutils}/bin/mkdir -p $(${pkgs.coreutils}/bin/dirname ${escapedAgeKeyFile})
           # age-keygen sets 0600 by default, no need to chmod.
-          ${pkgs.age}/bin/age-keygen -o ${escapedAgeKeyFile}
+          ${pkgs.age}/bin/age-keygen -o ${escapedAgeKeyFile} ${lib.join " " cfg.age.extraGenerateKeyArgs}
```
LucasWagler: Potential footgun here!

This example config breaks on a fresh system:
```nix
sops.age = {
  keyFile = "/fileA";
  extraGenerateKeyArgs = [ "-o" "/fileB" ];
};
```
Only `fileB` is generated, but sops-nix looks at non-existent `fileA` and breaks.

## Solution Idea:
```suggestion
          ${pkgs.age}/bin/age-keygen ${lib.join " " cfg.age.extraGenerateKeyArgs} -o ${escapedAgeKeyFile}
```
Also, add an assertion preventing "-o" in `extraArgs`.

(*Maybe* add a `let ageKeygenArg = "-o"` so that it is shared between the assertion & the script, for the unlikely case that a future `age-keygen` version changes the argument name or syntax.)

The downside is that `extraArgs` no longer overrides *everything*, which may confuse some users or impact a niche case. If `age-keygen` ever does change "-o" to represent something else, there will be a period in which users can't use `extraArgs` as an escape hatch for the new meaning of "-o", so they'll be stuck waiting for a `sops-nix` update.

LucasWagler: Alternative solution: leave `extraArgs` overriding everything, but add a warning if "-o" is present. This is a weaker guard that allows breakage, though.

SuperSandro2000:
```suggestion
          ${pkgs.age}/bin/age-keygen -o ${escapedAgeKeyFile} ${escapeShellArgs.join cfg.age.extraGenerateKeyArgs}
```

modules/nix-darwin/default.nix
```diff
@@ -173,7 +173,7 @@ let
             echo generating machine-specific age key...
             mkdir -p "$(dirname ${escapedKeyFile})"
             # age-keygen sets 0600 by default, no need to chmod.
-            ${pkgs.age}/bin/age-keygen -o ${escapedKeyFile}
+            ${pkgs.age}/bin/age-keygen -o ${escapedKeyFile} ${lib.join " " cfg.age.extraGenerateKeyArgs}
```
SuperSandro2000:
```suggestion
            ${pkgs.age}/bin/age-keygen -o ${escapedKeyFile} ${lib.escapeShellArgs cfg.age.extraGenerateKeyArgs}
```

modules/sops/default.nix
```diff
@@ -511,7 +528,7 @@ in
                 echo generating machine-specific age key...
                 mkdir -p $(dirname ${escapedKeyFile})
                 # age-keygen sets 0600 by default, no need to chmod.
-                ${pkgs.age}/bin/age-keygen -o ${escapedKeyFile}
+                ${pkgs.age}/bin/age-keygen -o ${escapedKeyFile} ${lib.join " " cfg.age.extraGenerateKeyArgs}
```
SuperSandro2000:
```suggestion
                ${pkgs.age}/bin/age-keygen -o ${escapedKeyFile} ${lib.escapeShellArgs cfg.age.extraGenerateKeyArgs}
```

PR #971, opened 2026-08-02
Fix tests by migrating away from postDeviceCommmands

eyJhb:
Utilise activationScripts with a dependency on setupSecrets, instead of the deprecated postDeviceCommands.
I tried to keep the changes as minimal as I could, and there are possibly other ways to fix this as well, but this felt like the most non-invasive change :)

I considered using `lib.mkBefore`, but it didn't feel entirely right when `deps` can be used instead.

### DIFF
--- a/checks/nixos-test.nix
+++ b/checks/nixos-test.nix
@@ -93,10 +93,11 @@ in
         };
 
         # must run before sops sets up keys
-        boot.initrd.postDeviceCommands = ''
+        system.activationScripts.earlySecrets.text = ''
           cp -r ${testAssets + "/age-keys.txt"} /run/age-keys.txt
           chmod -R 700 /run/age-keys.txt
         '';
+        system.activationScripts.setupSecrets.deps = [ "earlySecrets" ];
 
         specialisation.pruning.configuration.sops.keepGenerations = 10;
       };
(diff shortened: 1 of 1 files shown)


checks/nixos-test.nix
```diff
@@ -93,10 +93,11 @@ in
         };
 
         # must run before sops sets up keys
-        boot.initrd.postDeviceCommands = ''
+        system.activationScripts.earlySecrets.text = ''
           cp -r ${testAssets + "/age-keys.txt"} /run/age-keys.txt
           chmod -R 700 /run/age-keys.txt
         '';
+        system.activationScripts.setupSecrets.deps = [ "earlySecrets" ];
```
SuperSandro2000, line 100:
```suggestion
        system.activationScripts.earlySecrets = {
          deps = [ "earlySecrets" ];
          text = ''
            cp -r ${testAssets + "/age-keys.txt"} /run/age-keys.txt
            chmod -R 700 /run/age-keys.txt
          '';
        };
```

SuperSandro2000 (commented):
otherwise fine for me

checks/nixos-test.nix
eyJhb: It can't be collapsed like that, as I set the dependency on `setupSecrets` and not `earlySecrets`. I could modify all instances to look like:

```nix
system.activationScripts = {
  setupSecrets.deps = [ "earlySecrets" ];
  earlySecrets.text = ''
    cp -r ${testAssets + "/age-keys.txt"} /run/age-keys.txt
    chmod -R 700 /run/age-keys.txt
  '';
};
```

Would that be better?

EDIT: But it might make it more clear to preserve the order, and keep them separate.

SuperSandro2000 (approved):

PR #973, opened 2026-08-09
sops-install-secrets: skip invalid secrets instead of aborting the whole run

### LINKED ISSUE #972: sops-install-secrets: one secret with an unknown owner aborts installation of all secrets
## Summary

`sops-install-secrets` validates the entire manifest up front and aborts on the **first** error, before mounting the secrets fs or decrypting anything. As a result, a single secret with an `owner` that does not resolve to an existing user causes **zero** secrets to be installed — `/run/secrets.d` is never even created.

A cosmetic misconfiguration in one unrelated secret therefore takes down every secret-consuming service on the machine.

## Observed

On a NixOS host, one secret was declared with `owner = "grmpf"` while that user was not declared on that particular machine (`users.mutableUsers = false`, so it could never appear). Activation output:

```
setting up secrets...
sops-install-secrets: manifest is not valid: failed to lookup user 'grmpf': user: unknown user grmpf
```

None of the 10 secrets in the manifest were installed. `/run/secrets.d` did not exist, `/run/secrets` was never symlinked. The age key (`/var/lib/sops-nix/key.txt`) was present and valid, and all encrypted source files existed in the store — 9 of the 10 secrets were `owner = "root"` and had nothing wrong with them.

Downstream fallout on that host, all from the one bad owner:

- `yggdrasil.service` — `243/CREDENTIALS`, `LoadCredential=` source missing → node off the overlay network entirely (the symptom that started the investigation)
- `systemd-networkd.service` — `243/CREDENTIALS` (wireguard private key credential), plus its four sockets
- `hyprspace.service`, `p2p-ssh-iroh.service` — crash-looping, hundreds of restarts
- `zerotier` — survived only because it fell back to a pre-existing on-disk identity; its init script had already logged `identity secret has changed, backing up old identity`
- `sshd-keygen` generated a throwaway host key into the missing secret's parent path, creating `/run/secrets` as a **real directory**, which will then also block `sops-install-secrets` from placing its symlink on a subsequent run

The machine was left reachable only by luck. On a headless host with `owner` pointing at a user that is missing for any reason, this is a remote lockout.

## Root cause

`validateManifest` returns on the first failing secret:

https://github.com/Mic92/sops-nix/blob/f1406619a3884cd5c47992a70b8b35c9c0fcb4c9/pkgs/sops-install-secrets/main.go#L731-L735

```go
for i := range m.Secrets {
	secret := &m.Secrets[i]
	if err := app.validateSecret(secret); err != nil {
		return err
	}
	...
}
```

`validateSecret` → `validateOwner` does a hard `user.Lookup`:

https://github.com/Mic92/sops-nix/blob/f1406619a3884cd5c47992a70b8b35c9c0fcb4c9/pkgs/sops-install-secrets/main.go#L574-L584

And this all happens before any installation work — `validateManifest` is called at `main.go:1305`, whereas `MountSecretFs` and the decryption/install path only start at `main.go:1325`:

https://github.com/Mic92/sops-nix/blob/f1406619a3884cd5c47992a70b8b35c9c0fcb4c9/pkgs/sops-install-secrets/main.go#L1305-L1327

So the failure mode is strictly all-or-nothing, and the "nothing" case is silent apart from one activation log line.

## Expected behaviour

Install as many secrets as possible. A per-secret problem should degrade only that secret.

Concretely, my suggestion:

1. In the secret loop, collect per-secret errors instead of returning the first one. Skip the secrets that failed validation, install the rest.
2. Log each skipped secret loudly (name + reason).
3. Still exit non-zero at the end, with all collected errors aggregated (`errors.Join`), so activation still reports failure and CI/deploys still notice — but the machine keeps working.

Errors that are genuinely manifest-global (mutually exclusive `gnupgHome`/`sshKeyPaths`/`ageKeyFile`, unreadable/undecryptable sops files affecting everything, template placeholder resolution) can reasonably stay fatal. It's the per-secret metadata validation — `owner`, `group`, `mode`, format — that should be isolated.

An unresolvable `owner` in particular has an obvious safe degradation even short of skipping: fall back to `owner = root` (i.e. the same treatment `-ignore-passwd` already applies at `main.go:605`), warn, and install the secret anyway. The file lands with tighter-than-requested ownership, which is fail-safe — no secret is exposed to anyone who was not already root. That would have kept every one of the 10 secrets, including the 9 that were pure `root`, in place.

## Additional suggestion: catch it at eval time

This class of mistake is also detectable in the NixOS module, which would be strictly better than any runtime handling. There is already an implicit coupling — `group` defaults to `config.users.users.${owner}.group`:

https://github.com/Mic92/sops-nix/blob/f1406619a3884cd5c47992a70b8b35c9c0fcb4c9/modules/sops/default.nix#L121-L124

but that only throws when `group` is left at its default. In my case `group` was set explicitly (`group = "root"`), so eval sailed through and the problem surfaced only at activation on the target machine.

An assertion alongside the existing owner/uid ones at `modules/sops/default.nix:449-459`, roughly:

```nix
{
  assertion = secret.owner != null -> config.users.users ? ${secret.owner};
  message = "sops.secrets.${secret.name}.owner refers to user '${secret.owner}', which is not declared in users.users";
}
```

would turn this into a build-time error. Note this only helps for statically declared users, so it complements rather than replaces the runtime robustness above (users created outside `users.users` would need an escape hatch, e.g. skipping the assertion when `users.mutableUsers` is true).

## Environment

- sops-nix rev `f1406619a3884cd5c47992a70b8b35c9c0fcb4c9`
- nixpkgs 26.11 (`26.11.20260804.e72e4f2`)
- NixOS, systemd 261.1, x86_64-linux
- Secrets declared indirectly via [clan-core](https://git.clan.lol/clan/clan-core) `clan.core.vars`, which generates `sops.secrets` with explicit `owner`/`group`/`mode` — worth mentioning because generator-driven setups make it easy for an `owner` to be templated onto a machine that doesn't declare that user.

DavHau:
Fixes #972.

`validateManifest` returned on the first bad secret, before anything was decrypted, so one secret with an unresolvable `owner` left the machine with no secrets at all.

Now such secrets are skipped individually, the rest are installed, and the collected errors are returned at the end — activation still fails, but the box keeps working. Check modes (`-check-mode=`) stay strict, since those are build-time gates.

Test added: fails on master (the healthy secret is missing), passes here.

### DIFF
--- a/pkgs/sops-install-secrets/main.go
+++ b/pkgs/sops-install-secrets/main.go
@@ -156,6 +156,16 @@ type appContext struct {
 	secretByPlaceholder map[string]*secret
 	checkMode           CheckMode
 	ignorePasswd        bool
+	// Secrets dropped by validation, reported after the healthy ones are in.
+	skipped []error
+}
+
+func (app *appContext) skippedError() error {
+	if len(app.skipped) == 0 {
+		return nil
+	}
+	return fmt.Errorf("%d of %d secrets were skipped: %w",
+		len(app.skipped), len(app.skipped)+len(app.manifest.Secrets), errors.Join(app.skipped...))
 }
 
 // Keep this in sync with `modules/sops/templates/default.nix`
@@ -729,15 +739,26 @@ func (app *appContext) validateManifest() error {
 		}
 	}
 
+	// A broken secret only disables itself; the rest are still installed and
+	// the collected errors are reported once everything else is done.
+	kept := 0
 	for i := range m.Secrets {
 		secret := &m.Secrets[i]
 		if err := app.validateSecret(secret); err != nil {
-			return err
+			fmt.Fprintf(os.Stderr, "skipping secret '%s': %s\n", secret.Name, err)
+			app.skipped = append(app.skipped, err)
+			continue
 		}
+		m.Secrets[kept] = *secret
+		kept++
+	}
+	m.Secrets = m.Secrets[:kept]
 
-		// The Nix module only defines placeholders for secrets if there are
-		// templates.
-		if len(m.Templates) > 0 {
+	// The Nix module only defines placeholders for secrets if there are
+	// templates.
+	if len(m.Templates) > 0 {
+		for i := range m.Secrets {
+			secret := &m.Secrets[i]
 			placeholder, present := m.PlaceholderBySecretName[secret.Name]
 			if !present {
 				return fmt.Errorf("placeholder for %s not found in manifest", secret.Name)
@@ -1361,8 +1382,14 @@ func installSecrets(args []string) error {
 	if err = app.validateManifest(); err != nil {
 		return fmt.Errorf("manifest is not valid: %w", err)
 	}
+	// validateManifest drops the secrets it rejected.
+	manifest.Secrets = app.manifest.Secrets
 
+	// Checks are build-time gates, so nothing is tolerated there.
 	if app.checkMode != Off {
+		if err := app.skippedError(); err != nil {
+			return fmt.Errorf("manifest is not valid: %w", err)
+		}
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


DavHau:
test failures seem unrelated to this change

PR #984, opened 2026-09-16
sops-install-secrets: bump to buildGoModule

### LINKED ISSUE #983: sops-install-secrets: buildGo125Module was removed from nixpkgs, evaluation of sops.package fails
### Summary

`sops-install-secrets` is built with the versioned builder `buildGo125Module`. nixpkgs turned that attribute into a `throw` on 2026-09-15 (Go 1.25 end-of-life), so on nixpkgs at or after that commit the default `sops.package` can no longer be evaluated. This breaks the NixOS module and the home-manager module for consumers whose nixpkgs contains that change.

### Error

```
error: Go 1.25 is end-of-life, and 'buildGo125Module' has been removed. Please use a newer builder version.
```

It appears while evaluating `home-manager.users.<user>.sops.package`, reached from `home-manager.users.<user>.systemd.user.services.sops-nix.Service.ExecStart` → `home-manager.users.<user>.xdg.configFile."systemd/user/sops-nix.service".source` → `systemd.services.home-manager-<user>.serviceConfig`. On the NixOS side the same error appears for `sops.package` through the `sops-install-secrets` unit. The failure happens during evaluation, before anything is built.

### Root cause

* `pkgs/sops-install-secrets/default.nix:3,8` uses `buildGo125Module`.
* NixOS/nixpkgs#563177 replaced that attribute in `pkgs/top-level/aliases.nix` with a `throw`.
* `sops.package` is not built against this repo's pinned nixpkgs: `modules/sops/default.nix:274` (NixOS) and `modules/home-manager/sops.nix:233` (home-manager) default it to `(pkgs.callPackage ../.. { }).sops-install-secrets`, so `buildGo125Module` is resolved against the consumer's nixpkgs at evaluation time.

### Reproduced with

* nixpkgs `be3dab5b1a3c41ed2db86093b9f4e048305630b7` (`master`)
* sops-nix `13616fff713a9f94055c66f15687ebdc17a335df` (current `master`, 2026-09-09)

Trigger: evaluating a NixOS or home-manager configuration that defines at least one `sops.secrets` entry, e.g. `nh os switch .` (equivalently `nixos-rebuild switch`). Evaluation of `config.system.build.toplevel` aborts.

### Impact

* Consumers tracking `github:Mic92/sops-nix/master` together with nixpkgs `master` cannot evaluate their configuration at all. `nixos-unstable` will behave the same way once it advances past #563177.
* The repo's checks use the pinned nixpkgs input (with a separate `nixpkgs-stable` input whose checks are suffixed `-25_11`), so CI does not surface this; it only affects users on newer nixpkgs.

c2fc2f:
close https://github.com/Mic92/sops-nix/issues/983


NickCao:
We should probably use the unversioned buildGoModule so this doesn't happen again.

c2fc2f:
> We should probably use the unversioned buildGoModule so this doesn't happen again.

K900 told me on the Matrix server:
> There's a reason we keep old versions around for a while

So that's why I pinned a specific version

c2fc2f:
A few months ago, @SuperSandro2000 (https://github.com/Mic92/sops-nix/commit/5c23dc81a7b13890e9d9e81b39ff1bafa84b5637) had also bumped the 1.25 version without using the unversioned version, so I followed the "standard."

pkgs/sops-install-secrets/default.nix
```diff
@@ -1,11 +1,11 @@
 {
   lib,
-  buildGo125Module,
+  buildGo126Module,
   stdenv,
   vendorHash,
   go,
 }:
-buildGo125Module {
+buildGo126Module {
```
SuperSandro2000: Just omit the version number, or we forget this for the next version again

c2fc2f: So we need to be careful because `golangci-lint` (https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/go/golangci-lint/package.nix) specifies its version as a raw number; so if one day `buildGoModule` exceeds the version specified as a raw number, `sops-nix` will fail because there will be an incompatible version mismatch (due to `pkgs/lint.nix`). Perhaps we should ensure that `golangci-lint` isn't pinned to a specific version either

c2fc2f:
I kept the same branch name because, from what I've seen, people use it for their configurations, so to avoid breaking their configurations, and especially since it doesn't change anything for me

BeatLink:
Status guys?

ShadowRZ:
Ping @Mic92 for merging

BeatLink:
Lol so many people use sops nix for their builds I see.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/Mic92/sops-nix
 
-go 1.25.0
+go 1.26.0
```
DerDennisOP: by omiting the version number wont this break here next time anyway? @SuperSandro2000

picnoir (approved):

pkgs/sops-install-secrets/default.nix
SuperSandro2000: We figure something out when it is necessary.

auipga (approved):

